PR #16, opened 2026-03-25
Publish (but not release) built tarball every build.

jamesremuscat:
Tweak to the GH workflow to make the tarball we build downloadable from the action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,12 @@ jobs:
       - name: Build
         run: make
 
+      - name: Publish build tarball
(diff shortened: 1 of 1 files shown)


benarmston (approved):
There will potentially be a lot of artefacts being created and stored.  Do we need to worry about a retention policy?  Other than that this LGTM.

jamesremuscat:
> Do we need to worry about a retention policy?

GitHub's default is 90 days, configurable anywhere between 1 and 400 days (for a private repo). We can adjust if we find it necessary.

PR #23, opened 2026-03-26
Initial implementation of `flight howto`

### LINKED ISSUE #20: Implement `flight-howto` tool
`flight-howto` should allow the user to view a list of available howto documents, and to view a specific howto document. For now let's stick with the noninteractive mode of previous iterations of the tool:

- `flight howto list` should give a list of all available documents;
- `flight howto show <doc>` should render the specified document.

How the documents are specified is left as an exercise for the implementer (possibly a numeric index?)

The documents themselves should be written in Markdown, and we should render the Md nicely to the user (suggest [Glamour](https://github.com/charmbracelet/glamour) for the purpose).

Future enhancements may add an interactive mode to the tool but let's get the basics in first.

jamesremuscat:
This PR implements an initial version of the `flight howto` command.

`flight howto` lists and renders Markdown documents from the `${FLIGHT_ROOT}/usr/share/doc/howtos-enabled` directory and subdirectories, using CharmBracelet's [glamour](https://github.com/charmbracelet/glamour) library.

There are a couple of places where I've copypasta'd code from `flight-core`, mostly around error handling relating to `urfave/cli`, that might be worthwhile sharing at some point (somehow).

This is my first Go program so I've no idea what's idiomatic and what's idiotic :wink: 

There are no doubt plenty of possible enhancements but this works as an MVD I think.

Resolves #20.


flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Print(rendered)
+	return nil
+}
+
+func _print_dir_contents(dir_path string) error {
+	file, err := os.Open(dir_path)
```
benarmston: This function could be simplified by using [os.ReadDir](https://pkg.go.dev/os#ReadDir), which returns a slice of [fs.DirEntry](https://pkg.go.dev/io/fs#DirEntry).  That would remove the need to open each file.

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+	rendered, err := glamour.Render(string(markdown), "dark")
+	if err != nil {
+		return err
+	}
+
+	fmt.Print(rendered)
+	return nil
+}
+
+func _print_dir_contents(dir_path string) error {
```
benarmston: Idiomatic Go would call this `printDirContents(dirPath string)`.

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+	"path/filepath"
+	"strings"
+
+	"charm.land/glamour/v2"
+	"github.com/urfave/cli/v3"
+)
+
+// Wrapper around exec.ExitError that avoids the default handling by urfave/cli.
+// TODO deduplicate from flight-core/tools.go (both type and func below)
+type SilentExitError struct {
```
benarmston: I can't see any reason that howto would need this struct.

In core, we execute other programs can result in them exiting with a non-zero exit status.  urfave/cli by default is rather noisy when it processes them.  I expect that all of the programs that we start will have good error messages that indicate something went wrong so we don't need the extra noise from urfave/cli.

In howto, no processes are being executed, so there's no need for this.

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func main() {
+	cmd := &cli.Command{
+		Name:  "flight howto",
+		Usage: "List and display Flight User Suite documentation",
+		Commands: []*cli.Command{
+			{
+				Name:    "list",
+				Aliases: []string{"l"},
```
benarmston: I would imagine that `ls` is a common alias people will expect.  

```suggestion
				Aliases: []string{"ls", "l"},
```

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+				Usage:   "list available howtos",
+				Action:  list,
+			},
+			{
+				Name:      "show",
+				Aliases:   []string{"s"},
+				Usage:     "show a howto",
+				ArgsUsage: "<howto>",
+				Action:    show,
```
benarmston, line 46: Flight Core has a function to ensure that arguments are provided.  See https://github.com/concertim/flight-user-suite/blob/main/flight-core/main.go#L341.  That would help with error handling.

```bash
$ ./flight-howto show
2026/03/27 15:01:04 read /home/ben/projects/flight-user-suite/dist/opt/flight/usr/share/doc/howtos-enabled: is a directory
```

```bash
$ sudo FLIGHT_ROOT=$FLIGHT_ROOT ./flight tools enable
Incorrect Usage: missing argument tool
```

(I'll look at setting up a shared library we can use for our tools.)

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func show(ctx context.Context, cmd *cli.Command) error {
+	fullPath := filepath.Join(howtoDir, cmd.Args().First())
+	markdown, err := os.ReadFile(fullPath)
+	if err != nil {
+		return err
+	}
+
+	rendered, err := glamour.Render(string(markdown), "dark")
```
benarmston: We could change `"dark"` to `"auto"` to have it automatically select between a dark and light theme. I know that nowhere else does, but where its easy, I think we might as well.

```suggestion
	rendered, err := glamour.Render(string(markdown), "auto")
```

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
```
benarmston: Typically, when an error is returned it would be wrapped with additional context about that error.  If I change the glamour theme from "dark" to "dar", the following error is printed:

```bash
$ ./flight-howto show flight-job/custom-templates.md
2026/03/27 15:14:19 glamour: error reading file: open dar: no such file or directory
```

By adding `glamour: error reading file: ` to the error message, there is a fair amount more context than the default error message of `open dar: no such file or directory`.  Or at least that's the Go theory.

We should do similar for all of the errors that we're returning.  Something like this:

```suggestion
		return fmt.Errorf("rendering howto: %w", err)
```

flight-howto/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+func list(ctx context.Context, cmd *cli.Command) error {
+	return _print_dir_contents(howtoDir)
+}
+
+func show(ctx context.Context, cmd *cli.Command) error {
+	fullPath := filepath.Join(howtoDir, cmd.Args().First())
+	markdown, err := os.ReadFile(fullPath)
+	if err != nil {
+		return err
```
benarmston: Sometimes we'll want to change the error message significantly, not just prepend some context to it.  In that case we can create a new type that satisfies the `Error` interface and return that.

I did that in core when handling errors whilst executing a tool.  See https://github.com/concertim/flight-user-suite/blob/main/flight-core/tools.go#L28 and https://github.com/concertim/flight-user-suite/blob/main/flight-core/tools.go#L185.  I think it's worth doing something similar here.

In core the difference to the error message is:

```bash
$ sudo FLIGHT_ROOT=$FLIGHT_ROOT ./flight tools enable deskopt
chmod /home/ben/projects/flight-user-suite/dist/opt/flight/usr/lib/flight-core/flight-deskopt: no such file or directory
```

```bash
$ sudo FLIGHT_ROOT=$FLIGHT_ROOT ./flight tools enable deskopt
Unknown tool: deskopt
```

flight-howto/main.go
benarmston: Actually, I don't think we can.  You're using `charm.land/glamour/v2` which has changed the API for doing this.  Looks like `charm.land/glamour/v2` is the latest release replacing `github.com/charmbracelet/glamour`.

benarmston (changes requested):
The tool is looking pretty good.  There's lots of obvious enhancements but I'll avoid listing those.  Instead I've (mostly) limited my review to improvements on what's actually here.

flight-howto/main.go
jamesremuscat: I've written some code that glamour/lipgloss insists should work, but doesn't on my terminal. I'll push it and let you try it on yours...

jamesremuscat:
@benarmston I think I've addressed all your points, to varying degrees of success. This is ready for another look.

flight-howto/main.go
benarmston: It works fine on my terminal.

benarmston (approved):
I've pushed a commit to address the handling of light background issues.  The commit message has details, but basically, we were running it in slightly different ways and you uncovered a bug in `flight-core`.  Thanks!

PR #28, opened 2026-03-27
POC flight shell

benarmston:
This PR adds a basic shell from which other flight tools can be ran.

* It runs any enabled tool matching the glob `/opt/flight/usr/lib/flight-core/flight-*`.
* It has readline support for editing lines, history and tab completion.


flight-core/main.go
```diff
@@ -95,6 +95,17 @@ func main() {
 				},
 			},
 		},
+		Commands: []*cli.Command{
+			{
+				Name:        "shell",
+				Usage:       "Enter a shell-like sandbox for running Flight tools",
+				Description: wordwrap.String("Enter a shell-like sandbox for the 'flight' too.", maxTextWidth),
```
jamesremuscat: Typo:

```suggestion
				Description: wordwrap.String("Enter a shell-like sandbox for the 'flight' tool.", maxTextWidth),
```

jamesremuscat (changes requested):
Spotted a typo. Obviously, that means it's literally unusable in the current state 😉 

Tab-completion only works one level deep (i.e. for `des` it'll suggest `desktop` but for `desktop l` it won't suggest anything) but I guess that's one of many potential future enhancements.

PR #31, opened 2026-03-30
Improve modified Flight prompt.

### LINKED ISSUE #27: Improve the changes made to PS1
The changes made to `PS1` feel clunky.  We should give some thought to how they might be improved, and to how we might incorporate some orange.

The best candidate we have so far is:

```
<flight env> [root@login1 (mycluster1) ~]#
```

With probably both `<flight env>` and `(myluster1)` being in orange.

jamesremuscat:
- Shorten "Flight env active" to "Flight"
- Highlight it and cluster name in Concertim orange

image

We now fully replace the existing prompt, rather than adding bits to the existing one. This means we can guarantee the format of the final prompt. (Perhaps we want to be more accommodating to users' preferences though?)

We go to a little effort to make sure an empty/default cluster name doesn't result in a double space in the middle of the prompt.

I could have gone a step further and removed the need for the `augment-bash-prompt` Python script but I didn't.

Resolves #27.

### DIFF
--- a/flight-starter/opt/flight/etc/profile.d/02-prompt.sh
+++ b/flight-starter/opt/flight/etc/profile.d/02-prompt.sh
@@ -1,26 +1,18 @@
 export FLIGHT_ORIG_ENV_PS1="${PS1}"
 
-if [ "$PS1" = "\\s-\\v\\\$ " ]; then
-  # Prompt hasn't been set yet, give it a default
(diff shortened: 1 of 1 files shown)


benarmston (approved):
> Perhaps we want to be more accommodating to users' preferences though?

I suspect that's going to be a requirement, but let's leave it until it is.

LGTM.

PR #37, opened 2026-03-30
Parse vncserver output and display connection details

### LINKED ISSUE #33: Save the YAML output from libexec/vncserver to the session's metadata
When a session is created, metadata about it is saved to `~/.local/state/flight/desktop/sessions/<id>/metadata.yml`.

Additional metadata is written to STDOUT by `libexec/vncserver`, this should be parsed and added to the data saved in `metadata.yml`.

Example output from `vncserver` is below.  Everything between `<YAML>` and `</YAML>` should be parsed and saved.  As flight-desktop is likely to be the only tool parsing that output, changing it to something more friendly (looking at you Ruby symbols in YAML) is fine.

```
<YAML>
:host: "login1.ben.cluster.1.alces.network"
:display: "3"
:log: "/home/rocky/.local/state/flight/desktop/sessions/d13359f9-e4c0-4a95-911a-32d2231fb80e/session.log"
:pidfile: "/home/rocky/.local/state/flight/desktop/sessions/d13359f9-e4c0-4a95-911a-32d2231fb80e/vncserver.pid"
</YAML>
/bin/bash-c(/bin/bash /home/rocky/.local/state/flight/desktop/sessions/d13359f9-e4c0-4a95-911a-32d2231fb80e/session.sh  ; /opt/flight/usr/libexec/vncserver -sessiondir /home/rocky/.local/state/flight/desktop/sessions/d13359f9-e4c0-4a95-911a-32d2231fb80e -kill) >> '/home/rocky/.local/state/flight/desktop/sessions/d13359f9-e4c0-4a95-911a-32d2231fb80e/session.log' 2>&1 &
```

benarmston:
This PR parses and stores the output from the VNC server which includes the host and display number.  Now that we have access to the display number, the details printed after a session is started can include connection details.

image

Resolves #33


benarmston:
If the session is determined to be public:

image

jamesremuscat (approved):
I spotted a double space. I decided to just fix it rather than comment!

LGTM.

benarmston:
I'll merge this now.  @ColonelPanics if you have any comments on this wording leave a comment here and I'll get to it.

ColonelPanics:
I think we should lean towards using the documentation for connection and the public-facing check was always a bit flaky (and a lot more focussed on cloud-driven use cases) so I don't think it's needed.

My initial notes on some "how to connect" documents focus more on "can your _client_ reach this host" because that's more indicative than if the public internet can. 

Ultimately I'd say that the info could likely be simplified and point to reading the embedded docs for connection guidance. 

I'm also not sure that the "NOT SECURE" note is necessarily true, recent changes to the vncserver script in the old flight-desktop uses the TLS arg (and there's possibly more we could do to ensure a secure session).

benarmston:
Thanks for the feedback @ColonelPanics.  Does this look good to you?  If it does, we can improve formatting and colourify it down the line.

```
Starting a 'terminal' desktop session:

✅ Your terminal session is ready!

A new destop session (terminal.9) has been started. To connect, use the details
below:

1. Connection Address

Copy and paste the first address into your VNC viewer. If that doesn't work, try
one of the alternatives:

* Primary:       192.168.1.117:5915
* Alternative 1: vnc://ben:4ZXTNURH@192.168.1.117:5915
* Alternative 2: 192.168.1.117:15

2. Password

If prompted, use this password: 4ZXTNURH

3. Need Help?

For a step-by-step guide on how to connect, run: 'flight howto show flight-desktop'

4. Manage this session

To view details or stop this session later, you will need the Session ID:

    544cb3e1-b722-44ae-a3c1-8a7cbedd97ca

(Tip: Run 'flight desktop --help' to see management commands)

Started at: 31 Mar 26 13:16 BST
```

ColonelPanics:
Yep that looks much better! 

I'm a bit hesitant to refer to the docs as a "step-by-step guide", I think of it as advice and possible methods because connection possibilities will vary based on the HPC environment. 

Would be nice if we can interact with sessions by name instead of ID but I'll leave that feedback for another time (plus various other UX things could remove that need, tui for example!)

PR #39, opened 2026-03-31
Add `kill` command

### LINKED ISSUE #34: Stop an active desktop session
Write  a `flight desktop kill <id>` command that kills the session identified by `<id>`.

The session can be killed by running

```
vncserver -kill -sessiondir <session_dir>
```

Where `<session_dir>` is the directory `~/.local/state/flight/desktop/sessions/<id>/` and `vncserver` is our vintage perl script that can be found at `${FLIGHT_ROOT}/usr/libexec/vncserver`.

If the session is successfully killed, its session directory should be removed.

benarmston:
This PR adds a `kill <id>` command which will kill an active session by running `vncserver -kill -sessiondir ...` and if successful remove its session directory.

If the session cannot be loaded due to missing files or directories, log why and remove the session directory.

If the session cannot be loaded due to a bad metadata file, log why, attempt to kill the vncserver and if that's successful remove the session directory.  If there is a `vncserver.pid` file in the session directory, it may be possible to kill the session even if the `metadata.yml` file has become corrupted FSR.

image

### DIFF
--- a/flight-desktop/main.go
+++ b/flight-desktop/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -87,6 +88,7 @@ func main() {
 		},
 		Commands: []*cli.Command{
 			startCommand(),
+			killSessionCommand(),
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


jamesremuscat (approved):
LGTM.

PR #44, opened 2026-03-31
List and show sessions

### LINKED ISSUE #35: Add `flight desktop show <id>`
After a session has started, we print certain information about it including its connection details.  User's are likely to want to have access to this information after the fact.

Add a `flight desktop show <id>` command to show the same details that are printed after a session has started.

See #34 for details on how `<id>` can be matched to a session.

### LINKED ISSUE #36: Add `flight desktop list` command to show current desktop sessions
Making use of `flight desktop show <id>` (#35) is much easier if its possible to get a list of desktop sessions. Add the command

````
flight desktop list
```

For each session (that is each directory in ~/.local/state/flight/desktop/sessions/`) the following should be shown:

* Name
* Identity (UUID)
* Type
* Creation date
* Display port

benarmston:
This PR adds a command to list and show sessions.  The output for the list, show and start commands have all seen improvements.

### List

image

A session is shown as broken, it its `markdown.yml` file exists, but cannot be read FSR.

### Show

The show command includes some very brief data about the session, intended to allow the user to have confidence that they've found the right one, and then displays the same connection information as the start command.

image

### Start

The presentation of the output for the start command has changed:

image

### Graceful colour downsampling

If stdout is not a TTY, the colour codes are stripped.  Similarly, if a terminal has limited colour support the colour will be downsampled to a set it can handle.  See https://github.com/charmbracelet/lipgloss?tab=readme-ov-file#color-downsampling.

image

image

Resolves #36 
Resolves #35

### DIFF
--- a/flight-desktop/main.go
+++ b/flight-desktop/main.go
@@ -92,6 +92,7 @@ func main() {
 			startCommand(),
 			listSessionsCommand(),
 			killSessionCommand(),
+			showSessionCommand(),
 		},
 	}
 
--- a/flight-desktop/main.go
+++ b/flight-desktop/main.go
@@ -26,6 +26,7 @@ var (
 
 	progName     string = "flight desktop"
 	flightRoot   string = "/opt/flight"
+	termWidth    int    = 80
 	maxTextWidth int    = 80
 )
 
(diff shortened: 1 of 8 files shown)


jamesremuscat (approved):
LGTM.

Changes to make `flight howto show flight-desktop` work are tracked by #49 (otherwise we'll need a `.md` on the end of the command).

PR #52, opened 2026-03-31
Add avail command

### LINKED ISSUE #48: Add `flight desktop avail` command to list available session types
Add `flight desktop avail` command to display details about the available session types, e.g., gnome, terminal, etc..

There is basic information about each session type at `${FLIGHT_ROOT}/usr/lib/desktop/types/<type>/metadata.yml`.  Currently, it contains a name, summary and a URL.

We should display a table containing that information for each session type.  Lipgloss contains a [hyperlink method](https://github.com/charmbracelet/lipgloss?tab=readme-ov-file#hyperlinks), so we should use that for the URL so that terminals that support links can display the link.

There's also a TODO to dynamically list available types in help text (`session_start.go:16) which might be good to tackle at the same time.

Later work will add a `verify` command that checks if all of the session type's dependencies are installed.  When that is undertaken, we should add a "State" column that display whether the dependencies are installed.

benarmston:
This PR adds a `flight desktop avail` command; removes the hardcoding of available session types from the start command; and categorises the commands in the help output.

image

image

Resolves #48

### DIFF
--- a/flight-desktop/main.go
+++ b/flight-desktop/main.go
@@ -93,6 +93,7 @@ func main() {
 			listSessionsCommand(),
 			killSessionCommand(),
 			showSessionCommand(),
+			typeAvailCommand(),
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


jamesremuscat (approved):
LGTM.

PR #53, opened 2026-03-31
Add spinner for start and kill commands

benarmston:
I've chosen a spinner package that doesn't have any dependencies.  There are lots of alternatives, if we need features that this one doesn't have, such as support for true colour.

When everything works:

[Screencast from 01-04-26 12:04:54.webm](https://github.com/user-attachments/assets/63a2e500-a8ab-47a5-b109-4d7320b7551d)

When it doesn't:

[Screencast from 01-04-26 12:07:42.webm](https://github.com/user-attachments/assets/e675ba64-f776-4a60-9b00-b925815065ea)


jamesremuscat (approved):
Image

PR #54, opened 2026-03-31
Check for VNC server process rather than assume

benarmston:
When a session is loaded, we check if the session's pidfile points to an active process, if it does, we consider the session to be active.  If there are any errors, we consider the session to be exited.

The show command includes the session's state and no longer displays the connection info for inactive sessions.

image

Refs #36


benarmston:
Merging without review.

PR #55, opened 2026-04-01
Improve flight-howto help

### LINKED ISSUE #30: Improve flight howto `--help` output
The current output for `flight howto --help` could do with some improvement.

There's likely some wording from https://github.com/openflighthpc/flight-howto/blob/master/lib/flight-howto/cli.rb and https://github.com/openflighthpc/flight-howto/tree/master/lib/flight-howto/commands that is worth incorporating.

It would also be nice to orangify it: see https://github.com/concertim/flight-user-suite/blob/main/flight-core/main.go#L108 for how flight-core does that.

```bash
# Build all of the different projects/tools etc.
$ make

# Point FLIGHT_ROOT at what we've just built.
$ export FLIGHT_ROOT=`pwd`/dist/opt/flight

# Change to the flight-howto directory and run flight-howto.  Using `go run ./...` will pickup the latest changes to the source code.
$ cd flight-howto
$ go run ./... --help
NAME:
   flight howto - List and display Flight User Suite documentation

USAGE:
   flight howto [global options] [command [command options]]

COMMANDS:
   list, l, ls  list available howtos
   show, s      show a howto
   help, h      Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h  show help
```

megan-godfrey:
This PR updates the wording and increases the orange-ness of the help text for `flight-howto`.

The custom help printer function has also been moved to a new package (in `flight-core/pkg`), which is now shared by `flight-core`, `flight-desktop` and `flight-howto`.

Resolves #30


flight-core/pkg/help_printer.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"io"
+	"log"
+	"regexp"
+
+	"github.com/cyucelen/marker"
+	"github.com/fatih/color"
+)
+
+func OrangifiedHelpPrinter(origHelpPrinter func(w io.Writer, templ string, data any)) func(w io.Writer, templ string, data any) {
```
benarmston: urfave/cli defines a type for the help printer:

```go
// HelpPrinterFunc prints help for the Command.
type HelpPrinterFunc func(w io.Writer, templ string, data any)
```

We can use this to make the signature for `OrangifiedHelpPrinter` a bit clearer.  The imports will need adjusting, so probably best to manually apply rather than accept this suggestion.

```suggestion
func OrangifiedHelpPrinter(origHelpPrinter cli.HelpPrinterFunc) cli.HelpPrinterFunc {
```

flight-core/pkg/help_printer.go
```diff
@@ -0,0 +1,30 @@
...
+	return func(w io.Writer, templ string, data any) {
```
benarmston, line 16: And this too.

```suggestion
	return cli.HelpPrinterFunc {
```

flight-howto/main.go
```diff
@@ -30,25 +31,29 @@ func init() {
 func main() {
 	cmd := &cli.Command{
 		Name:  "flight howto",
-		Usage: "List and display Flight User Suite documentation",
+		Usage: "View user guides for your HPC environment",
```
benarmston: Both core and desktop provide a description. Do you think one is needed for howto?

benarmston (changes requested):
The help output for howto looks much nicer. Thanks.

I've added a couple of suggestions for making the orangified help printer a little easier to read.  Applying them will won't work as the imports will not be correct, but having the diff in the comment seemed useful.

The desktop and core commands also have a description set for their top-level command and all subcommands. Do you think there is something sensible that could be added for the howto commands?  If you think not, that's fine with me.

flight-core/pkg/help_printer.go
megan-godfrey: I don't really understand why but making this change didn't work

flight-howto/main.go
megan-godfrey: Ideally I think the `Usage` info would be shown in the `description` section instead of the `name` section for all of the apps. For now I can add a description section to the howto help for consistency?

flight-core/pkg/help_printer.go
benarmston: Quite right, this was a bad suggestion.  The suggestion for the line above is still good though IMO.

flight-core/pkg/styling_helper.go
```diff
@@ -28,3 +36,12 @@ func OrangifiedHelpPrinter(origHelpPrinter func(w io.Writer, templ string, data
 		fmt.Fprint(w, out)
 	}
 }
+
+func WrappedString(s string) string {
+	termWidth, _, err := term.GetSize(int(os.Stdout.Fd()))
+	if err != nil {
+		termWidth = maxTextWidth
+	}
```
benarmston: This is going to run every time the function is called.  I'd expect the terminal width not to change for a single run of a CLI program, so we only need to calculate this once.  `flight-desktop` was previously using an `init` function to do so.  It's worth checking to see if we can use the same trick in this package.

Once we add a TUI, we should support the terminal being resized, but bubbletea provides its own way of doing that.

benarmston (commented):
LGTM. Once the merge conflict is resolved, this is good to go.

flight-core/pkg/styling_helper.go
benarmston: Ignore this. It was queued up and I forgot to remove it prior to submitting my review.

Issue #46, opened 2026-03-31
[Flight Desktop] Use names instead of session IDs


jamesremuscat:
We allow users to specify a name for their VNC sessions. We should allow them to use this name in place of session ID in places where a session ID is required.

benarmston:
Currently, the name is optional and nothing prevents sessions from having duplicate names.  This makes using names as identifiers difficult to say the least.

What do you think to replacing session IDs with names?  That is, instead of generating a UUID and using that as name of the session directory, we would use the session name as the name of the session directory.  So, `flight desktop start terminal --name nethack` would generate the session directory `~/.local/state/flight/desktop/sessions/nethack`.

We'd need to ensure that the name wasn't already taken,which should be easy and quick.  We could either make the name mandatory or have it optional and generate a random name based on the session type, e.g., `terminal.dAJjqK`.

User would likely reuse session names over time, and the above suggestion has no unique-over-time identifier for a session.  I can't imagine that would be a problem.

jamesremuscat:
> nothing prevents sessions from having duplicate names

Yes, implicit in using names in this way is that they should be constrained to be unique!

Replacing IDs with names is a great shout - if names are unique (at any given instant) then there's no need to have separate IDs, that's just complication.

Names should remain optional, and we should generate a name if one isn't specified. (I still like `appropriately-humorous-wordplay` style names, but something random based on the session type might have more utility...)

PR #56, opened 2026-04-01
Better handle `list` column widths

benarmston:
The primary goal is to ensure that the table is no wider than the width of the terminal, otherwise we "lose" the ID and cannot copy paste from the list output to show or kill a session.

To achieve this there needs to be some flexibility in the column widths, so some have a width set whilst others have a max width.  After a small amount of playing with this seems the best compromise.

On a wide terminal (130 columns):

image

On a "standard" terminal (80 columns):

image

It probably doesn't work below 80 columns :man_shrugging:.

We could probably do better by examining the width of the terminal, but that seems like a lot of work right now.  If we use session names as the session IDs (https://github.com/concertim/flight-user-suite/issues/46#issuecomment-4164363028), we'll have one less column to worry about, which might help.  Or just shorter IDs in general.

### DIFF
--- a/flight-desktop/session_list.go
+++ b/flight-desktop/session_list.go
@@ -35,6 +35,12 @@ func listSessionsCommand() *cli.Command {
 }
 
 func sessionsTable(sessions []*Session) error {
+	namecolWidth := 8
+	maxNamecolWidth := 20
+	typecolWidth := 8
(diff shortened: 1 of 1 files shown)


benarmston:
Merging without review.

PR #58, opened 2026-04-01
Disable all tools except `flight-howto` in distribution.

### LINKED ISSUE #25: Enable the howto tool by default; disable other tools by default
When the building the tar file, we should ensure that the howto tool is executable for owner, group and other.  This likely going to be the default behaviour assuming an implementation in Go, but let's have this issue to ensure that is the case.

jamesremuscat:
This seemed like the best place to do it...

`flight-howto` doesn't itself have any howto documents related to it, or otherwise we'd also need to worry about symlinking those into the appropriate directory.

Resolves #25.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ $(MODULES):
 
 $(DIST): $(MODULES)
 	rsync -rlptgo $(foreach module,$(MODULES),$(module)/dist/) $(DIST)/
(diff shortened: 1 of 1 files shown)


benarmston (approved):
This LGTM. It's worth noting that if we ever want a tool to be enabled by default and it also has howto guides, we'll need to create the symlinks for them here too.

PR #59, opened 2026-04-01
Add doctor command to perform system health check

### LINKED ISSUE #57: Add a `flight desktop doctor` command to check for prerequisites
Add `flight desktop doctor` command to check for prerequisites and report.

`flight-desktop` itself should maintain a list of required dependencies, e.g., `vncserver`, `vncpasswd`, etc..  Each desktop type should also maintain a list of its required programs, e.g., `xterm`, `xrdb`, etc..

A dependency is one of two things:

1. An executable that needs to exist, either somewhere on `PATH` (e.g., `vncpasswd`) or at the exact location given (e.g., `/usr/bin/vncserver`). 2. A directory that must not be empty, e.g., `/usr/share/fonts`.

The `doctor` command should check that each executable is available (see https://pkg.go.dev/os/exec#LookPath) and report which are present and which are missing.  Similarly, it should check that each directory exists and is not empty.

The report should be separated into "Core requirements" and requirements for each desktop type, e.g., "Gnome desktop type requirements".

If there are any requirements that cannot be found, the command should exit non-zero.  If they are all found its should exit zero.

### Stretch goals

* Add a `state` attribute to desktop types to record if they are verified or not.
* Have `flight desktop doctor` manage that attribute. I.e., it sets it to `verified` or `unverified` after each time it is ran.
* Display this state in the `avail` command.
* If an attempt to launch an unverified desktop type is made, abort early with a message to contact the cluster sysadmin.

### Future enhancements

Allow `flight-desktop` itself and each desktop type to list optional requirements.  The report should list whether they can be found and what improved functionality they provide, e.g., `apg` provides nicer passwords.

Allow desktop types to provide an arbitrary script to run to check its dependencies are satisfied.

benarmston:
Flight desktop, itself, and the desktop types are able to specify any dependencies they have. It is expected that desktop, itself will list core dependencies that are common to all desktop types, and the individual types list only their own dependencies.

There are two types of dependency, executables and directories.

An executable can be specified either as an absolute path, or as a name to be lookedup from PATH. In either case, the file must be found and must be executable.

A directory must exist and be non-empty to satisfy the dependency.

In both cases a list of paths any one of which can satisfy the dependency is provided.

Flight desktop does so in its new configuration file `${FLIGHT_ROOT}/etc/desktop.yml`.

The desktop types do so in their new dependency file `${FLIGHT_ROOT}/usr/lib/desktop/types/<type>/dependencies.yml`.

[Screencast from 01-04-26 16:06:11.webm](https://github.com/user-attachments/assets/1f083326-1d25-4b32-9f3b-50bd2e1f51f0)

Resolves #57

### DIFF
--- a/flight-desktop/config.go
+++ b/flight-desktop/config.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+type config struct {
+	Dependencies []dependency `yaml:"dependencies"`
+}
+
+type dependency struct {
+	Type  string   `yaml:"type"`
+	Paths []string `yaml:"paths"`
+}
+
+func loadConfig() (*config, error) {
+	path := filepath.Join(flightRoot, "etc", "desktop.yml")
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("loading config: %w", err)
+	}
+	var config config
+	err = yaml.Unmarshal(data, &config)
+	if err != nil {
+		return nil, fmt.Errorf("loading config from %s: %w", path, err)
+	}
+	return &config, nil
+}
--- a/flight-desktop/doctor.go
+++ b/flight-desktop/doctor.go
@@ -1,0 +1,192 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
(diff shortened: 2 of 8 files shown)


jamesremuscat:
Please can we document `doctor` in the `flight-desktop.md` howto?

benarmston:
> Please can we document `doctor` in the `flight-desktop.md` howto?

'Tis done.

jamesremuscat (approved):
LGTM.

PR #60, opened 2026-04-01
`.md` suffix is optional when using `flight howto show`.

### LINKED ISSUE #49: [flight howto] Assume `.md` extension and hide from users?
We like the look of commands such as `flight howto show flight-desktop` but these currently don't work with `flight howto`, which requires the `.md` extension.

All howtos will be Markdown documents, so we can:
- assume a `.md` extension in `flight howto show` if one is not supplied
- hide the extension in `flight howto list` to make it look cleaner

jamesremuscat:
Also hide this extension when using `flight howto list` - it's just visual noise since all the howtos are Markdown documents by definition.

Resolves #49.


benarmston (approved):
LGTM.

PR #66, opened 2026-04-02
Fix miscellaneous desktop issues

benarmston:
* Add support to `doctor` for optional dependencies; make `gconftool-2` optional.
* Use `gsettings` instead of `gconftool-2`; it is a required dependency.
* Improve output by adding a description for each dependency.
* Improve output when checking for non-empty directories; list all that satisfy.
* Prevent desktop type name and id from deviating by removing name.

image


benarmston:
@jamesremuscat I can't decide if the adjusted "Fonts" output is still misleading.  It's only listing those directories that we're checking for; however the gnome session would presumably be using any fonts that are on the system even if we're not checking for them.

What do you think to having the line be: 

```
Fonts : Available
```

or

```
Fonts : No fonts found in
        /usr/share/X11/fonts/
        /usr/share/fonts/
        /usr/share/fonts/X11/
```

jamesremuscat:
I think in the grand scheme of things, what we currently have is probably OK - though if a user has fonts in `~/.fonts` presumably Gnome will be happy to use those? Maybe this should be an optional dependency since we're checking for the most common configurations, not all possible ones.

benarmston:
> Maybe this should be an optional dependency since we're checking for the most common configurations, not all possible ones.

That might work, as we're likely to have false negatives and unlikely to have false positives.  So it could be displayed as a warning (not an error) as:

```
Fonts : Unable to find fonts in standard locations. Ensure you have fonts enabled.
```

PR #67, opened 2026-04-02
Replace session id with name

### LINKED ISSUE #46: [Flight Desktop] Use names instead of session IDs

benarmston:
This PR replaces session IDs with session names.

The session ID field is no longer present and the session name field is now mandatory.  If the user does not provide one, we generate a random name for them.  There are two generation strategies available, one that generates names such as `xterm.4VPSIB6K` and the other that generates names like `unimpressively-run-xterm`.

A configuration setting is available to select which type of name generation to use.

The session name needs to be unique per-user, this is guaranteed by using it as the session directory name.  All sessions for a user are in the same parent directory and creation is atomic, so if there is already a session there, the attempt to create the directory will fail and the user is informed.

Resolves #46


jamesremuscat:
When I first tried running this it gave me a gnarly stack trace:

```
<flight> [james@jrem-laptop flight-desktop]$ go run . start xterm
panic: invalid argument to IntN

goroutine 1 [running]:
math/rand/v2.(*Rand).IntN(...)
        /usr/local/go/src/math/rand/v2/rand.go:193
math/rand/v2.IntN(...)
        /usr/local/go/src/math/rand/v2/rand.go:313
main.sample[...]({0x0?, 0x31b0c315a4e0?, 0x10?})
        /home/james/workspace/flight-user-suite/flight-desktop/name_generation.go:75 +0x71
main.(*nameGenerator).Generate(0x31b0c2ffe180)
        /home/james/workspace/flight-user-suite/flight-desktop/name_generation.go:50 +0x85
main.startSessionCommand.func2({0x7455f8, 0x31b0c315a000}, 0x31b0c314edc8)
        /home/james/workspace/flight-user-suite/flight-desktop/session_start.go:64 +0xab
github.com/urfave/cli/v3.(*Command).run(0x31b0c314edc8, {0x7455f8, 0x31b0c315a000}, {0x31b0c3144f20, 0x2, 0x2})
        /home/james/go/pkg/mod/github.com/urfave/cli/v3@v3.8.0/command_run.go:364 +0x2e6c
github.com/urfave/cli/v3.(*Command).run(0x31b0c314f8c8, {0x7455f8, 0x31b0c2f3d8c0}, {0x31b0c2f3c150, 0x3, 0x3})
        /home/james/go/pkg/mod/github.com/urfave/cli/v3@v3.8.0/command_run.go:298 +0x26a5
github.com/urfave/cli/v3.(*Command).Run(...)
        /home/james/go/pkg/mod/github.com/urfave/cli/v3@v3.8.0/command_run.go:93
main.main()
        /home/james/workspace/flight-user-suite/flight-desktop/main.go:101 +0x3d3
exit status 2
```

I'd not yet re-run `make` at the top level so the required name lists weren't present in the appropriate place under `FLIGHT_ROOT`, though it looks like the code _should_ handle that gracefully... perhaps it was some other artefact of how I'm running it. Given that end users will only ever be running this from a distribution package, perhaps we don't need to care too much?

EDIT: Actually, maybe it was the lack of configuration (so `ng.config.Strategy` was not `absurd` and we never tried to load combinations before generating, at which point the strategy was also not `meaningless`)? Perhaps the configuration should have a default?

benarmston:
> EDIT: Actually, maybe it was the lack of configuration (so `ng.config.Strategy` was not `absurd` and we never tried to load combinations before generating, at which point the strategy was also not `meaningless`)? Perhaps the configuration should have a default?

I've added default to `config.go`, to handle that part of the configuration not being present.

jamesremuscat (changes requested):
Thanks for that config fix - we also need to update `flight-desktop/opt/flight/usr/share/doc/flight-desktop/flight-desktop.md` to refer to names rather than IDs, then this can be merged.

benarmston:
> Thanks for that config fix - we also need to update `flight-desktop/opt/flight/usr/share/doc/flight-desktop/flight-desktop.md` to refer to names rather than IDs, then this can be merged.

Good catch. Thank you.

PR #68, opened 2026-04-02
Run `doctor` checks before launching a desktop session.

### LINKED ISSUE #62: [Flight Desktop] Don't try to create sessions if `doctor` says we can't
If `flight desktop doctor` reports that it's not possible to launch a particular type of session, we shouldn't attempt to launch it when `flight desktop start` is invoked.

jamesremuscat:
This PR adds in a "checking dependencies" step before attempting to launch a desktop session.

If a dependency is missing, we report it:
image
(Note: since this screenshot was taken I've removed the backticks to be consistent with `flight desktop doctor`'s output)

We then abort without attempting to launch the session. Otherwise we report that everything is OK:
image

Resolves #62.

### DIFF
--- a/flight-desktop/session_start.go
+++ b/flight-desktop/session_start.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/muesli/reflow/wordwrap"
@@ -104,6 +105,9 @@ func checkDependencies(ctx context.Context, sessionType string) (bool, error) {
 	p := createPin("Checking system dependencies...")
 	cancel := p.Start(ctx)
 	defer cancel()
+
+	// Add a small delay to stop the spinner from flickering
+	<-time.After(1 * time.Second)
 
 	globalResults, globalDepsOK := runDoctor(requiredDependencies(config.Dependencies))
 
--- a/flight-desktop/session_start.go
+++ b/flight-desktop/session_start.go
@@ -44,13 +44,13 @@ func startSessionCommand() *cli.Command {
 			sessionType := cmd.StringArg("type")
 			fmt.Printf("Starting a '%s' desktop session:\n\n", sessionType)
 
-			p := pin.New("Starting session...",
-				pin.WithSpinnerColor(pin.ColorCyan),
-				pin.WithTextColor(pin.ColorGreen),
-				pin.WithDoneSymbol('\u2705'),
-				pin.WithFailSymbol('\u274c'),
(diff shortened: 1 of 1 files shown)


flight-desktop/session_start.go
```diff
@@ -98,3 +98,52 @@ func assertTypeValid(argName string, argIndex int) cli.BeforeFunc {
 		return ctx, nil
 	}
 }
+
+func checkDependencies(ctx context.Context, sessionType string) (bool, error) {
+	config, err := loadConfig()
+	p := createPin("Checking system dependencies...")
+	cancel := p.Start(ctx)
+	defer cancel()
```
benarmston, line 107: The checking of dependencies is so fast that there is a very brief flicker of a spinner before it changes to something else.  What do you think to either, not display the spinner, or adding a delay with `<- time.After(1*time.Second)`?

flight-desktop/session_start.go
```diff
@@ -98,3 +98,52 @@ func assertTypeValid(argName string, argIndex int) cli.BeforeFunc {
...
+
+	globalResults, globalDepsOK := runDoctor(config.Dependencies)
```
benarmston: This is checking for both required and optional dependencies.  I'm not sure why we're checking for optional ones here and I suspect that's not your intention.  We now have a `requiredDependencies` function that filters a list of dependencies to only those that are required.

```suggestion
	globalResults, globalDepsOK := runDoctor(requiredDependencies(config.Dependencies))
```

flight-desktop/session_start.go
```diff
@@ -98,3 +98,52 @@ func assertTypeValid(argName string, argIndex int) cli.BeforeFunc {
...
+	if err != nil {
+		return false, err
+	}
+
+	err = sessionTypeDef.loadDependencies()
+	if err != nil {
+		return false, err
+	}
+
+	typeResults, typeDepsOK := runDoctor(sessionTypeDef.dependencies)
```
benarmston: Similar here.

```suggestion
	typeResults, typeDepsOK := runDoctor(requiredDependencies(sessionTypeDef.dependencies))
```

benarmston (changes requested):
A couple of comments below for you to look at please. Other than those this is looking good.

flight-desktop/session_start.go
jamesremuscat: I like having the spinner there (it's consistent with the "Starting session..." step) so I've opted for the delay.

jamesremuscat:
@benarmston I've addressed your comments.

benarmston (approved):
Thanks for those changes.  LGTM.

PR #69, opened 2026-04-02
Clean environment when launching VNC session

### LINKED ISSUE #61: [Flight Desktop] Multiple Gnome sessions for the same user don't work
Stu reports that it's not possible for a user to launch more than one Gnome session, possibly due to something systemd-y being user-scoped.

We should investigate and make this work.

benarmston:
The environment is cleaned to a configurable whitelist. The default whitelist is:

- DISPLAY
- FLIGHT_ROOT
- HOME
- LANG
- PATH
- PWD
- UID
- USER
- VNCDESKTOP

This is sufficient to allow multiple gnome sessions to be active at once.

Fixes #61

### DIFF
--- a/flight-desktop/config.go
+++ b/flight-desktop/config.go
@@ -10,6 +10,7 @@ import (
 
 type config struct {
 	Dependencies []dependency `yaml:"dependencies"`
+	EnvWhitelist []string     `yaml:"environment_whitelist"`
 }
 
 type dependency struct {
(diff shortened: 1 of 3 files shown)


benarmston:
@ColonelPanics if you want to test this out you can download the build for this PR at https://github.com/concertim/flight-user-suite/actions/runs/23913367037.

benarmston:
I've tested that this works correctly on a https://github.com/concertim/barely-a-cluster running on garycloud.  Going to merge without review.

PR #75, opened 2026-04-07
Improve session failure handling

### LINKED ISSUE #65: [Flight Desktop] `start` can spam empty directories if it fails
If `flight desktop start` fails to create a session, we should make sure it cleans up the session directory.

We previously had a `clean` command that mopped up such directories...

benarmston:
This PR deals with the issue of empty directories being spammed if launching a script fails.  It does so in three ways:

* If we can detect a failure when launching the desktop session, we clean up the session directory.
* A "cleaner" script is launched after the VNC server. This script waits for the VNC server process and when that process exits, cleans up the session's session directory.
* Added a `clean` command that will remove the session directory for any `exited` or `broken` sessions.

NOTE: The clean command doesn't handle remote sessions correctly, as they don't exist yet.  Once added, the clean command will need to be updated to skip them too.

Resolves #65

### DIFF
--- a/flight-desktop/session_clean.go
+++ b/flight-desktop/session_clean.go
@@ -52,7 +52,7 @@ You can specify which sessions are cleaned by providing the optional <id> parame
 			timer := time.After(1 * time.Second)
 
 			var firstErr error
-			skippedStyle := lipgloss.NewStyle().Foreground(primary).MarginLeft(1)
+			skippedStyle := lipgloss.NewStyle().Foreground(lightDark(primary, cream)).MarginLeft(1)
 			cleanedStyle := lipgloss.NewStyle().Foreground(lipgloss.Green).MarginLeft(1)
 			failedStyle := lipgloss.NewStyle().Foreground(lipgloss.Red).MarginLeft(1)
 			cleaned := make([]*Session, 0)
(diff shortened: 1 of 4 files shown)


flight-desktop/session_clean.go
```diff
@@ -0,0 +1,121 @@
...
+				pin.WithDoneSymbol('\u2705'),
+				pin.WithFailSymbol('\u274c'),
+				pin.WithFailColor(pin.ColorRed),
+			)
+			cancel := p.Start(ctx)
+			defer cancel()
+			timer := time.After(1 * time.Second)
+
+			var firstErr error
+			skippedStyle := lipgloss.NewStyle().Foreground(primary).MarginLeft(1)
```
jamesremuscat: There's a slight problem with this style (on my terminal at least)...

Image

I thought the output was broken at first! You can barely see the name of the skipped session on the bottom line.

jamesremuscat (changes requested):
This worked to clean up a dead session I had lying around - there's one cosmetic issue to address though!

benarmston:
The cosmetic issue is addressed and I've rebased on latest `main` to address `Session.ID` no longer being around.

PR #77, opened 2026-04-07
Include basic documentation in release.

### LINKED ISSUE #63: [Flight Core] Write content for `01-about-flight-user-suite.md`
`flight-core/opt/flight/usr/share/doc/flight-core/01-about-flight-user-suite.md` contains no content.

jamesremuscat:
The contents of `README.md` are based on the instructions in Flight Center for downloading and using FUS introduced in https://github.com/alces-software/alces-flight-center/pull/3839. I've made a few small modifications to reflect the different expected audience.

These instructions seem reasonably well-suited to serve both as a repo README and the on-cluster documentation - so the release process now copies `README.md` into the FUS howto directory, as the previously-empty file `01-about-flight-user-suite.md`.

We might want these to diverge in time, but for now, this seems like a double tick on the todo list.

Resolves #63.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,8 @@ $(DIST): $(MODULES)
 	chmod 0555 $(DIST)/opt/flight/usr/lib/flight-core/flight-howto
 # Note: if we ever want default-enabled tools that have documentation we will
 # also need to symlink the docs into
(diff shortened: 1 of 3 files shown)


jamesremuscat:
I don't see much value in a code review here.

PR #80, opened 2026-04-07
Add config commands to manage settings

benarmston:
The config commands allow setting per-user and cluster-wide settings. The settings can be listed, set and got.  If the `--global` flag is provided only globally set, i.e., cluster-wide, settings are acted on. Otherwise, listing and setting merges cluster-wide and per-user settings and set sets the per-user setting.

The only setting that can currently be managed is `autostart`. If set the flight environment is automatically started when a login shell starts.

Example session:

```sh
$ cd dist
$ export FLIGHT_ROOT=`pwd`/opt/flight
$ XDG_CONFIG_DIRS=${XDG_CONFIG_DIRS}:`pwd`/../dist/etc/xdg
$ ./opt/flight/bin/flight config list
No config settings have been set.
$ ./opt/flight/bin/flight config set autostart on --global
$ ./opt/flight/bin/flight config list
autostart=on
$ source ./etc/profile.d/zz-flight-starter.sh 
                                   __ _ _       _     _  ==>
   ==>                            / _| (_)     | |   | |  ==>
  ==>   ___   _ __    ___  _ __  | |_| |_  __ _| |__ | |_  ==>
 ==>   / _ \ | '_ \  / _ \| '_ \ |  _| | |/ _` | '_ \| __|  ==>
==>   | (_) || |_) ||  __/| | | || | | | | (_| | | | | |_    ==>
 ==>   \___/ | .__/  \___||_| |_||_| |_|_|\__, |_| |_|\__|  ==>
  ==>        |_|                           __/ |           ==>
   ==>  Welcome to your cluster           |___/           ==>
    ==>  Flight User Suite v0.0.2-31-gbbec88d-dirty
     ==>  Based on Ubuntu 24.04.4 LTS

Flight environment is now active.
<flight> [ben@bens-laptop dist]$ flight-stop
Flight environment is now inactive.
$ ./opt/flight/bin/flight config set autostart off
$ ./opt/flight/bin/flight config list
autostart=off
$ ./opt/flight/bin/flight config list --global
autostart=on
$ source ./etc/profile.d/zz-flight-starter.sh 
$ flight-start
                                   __ _ _       _     _  ==>
   ==>                            / _| (_)     | |   | |  ==>
  ==>   ___   _ __    ___  _ __  | |_| |_  __ _| |__ | |_  ==>
 ==>   / _ \ | '_ \  / _ \| '_ \ |  _| | |/ _` | '_ \| __|  ==>
==>   | (_) || |_) ||  __/| | | || | | | | (_| | | | | |_    ==>
 ==>   \___/ | .__/  \___||_| |_||_| |_|_|\__, |_| |_|\__|  ==>
  ==>        |_|                           __/ |           ==>
   ==>  Welcome to your cluster           |___/           ==>
    ==>  Flight User Suite v0.0.2-31-gbbec88d-dirty
     ==>  Based on Ubuntu 24.04.4 LTS

Flight environment is now active.
<flight> [ben@bens-laptop dist]$ flight-stop
Flight environment is now inactive.
```

### DIFF
--- a/flight-core/config_command.go
+++ b/flight-core/config_command.go
@@ -211,12 +211,16 @@ func saveConfig(key, value string, global bool) error {
 	for k, v := range config {
 		fmt.Fprintf(&b, "%s=%s\n", k, v)
 	}
-	return os.WriteFile(path, []byte(b.String()), 0o600)
+	if global {
+		return os.WriteFile(path, []byte(b.String()), 0o666)
+	} else {
+		return os.WriteFile(path, []byte(b.String()), 0o600)
+	}
 }
 
 func globalConfigPath() (string, error) {
 	name := filepath.Join("flight", "settings.config")
-	paths := xdg.ConfigDirs
+	paths := []string{"/etc/xdg"}
 	return configPath(name, paths)
 }
 
@@ -235,7 +239,7 @@ func configPath(name string, paths []string) (string, error) {
 		if exists(dir) {
 			return p, nil
 		}
-		if err := os.MkdirAll(dir, os.ModeDir|0o700); err == nil {
+		if err := os.MkdirAll(dir, os.ModeDir|0o777); err == nil {
 			return p, nil
 		}
 
--- a/flight-core/config_command.go
+++ b/flight-core/config_command.go
@@ -1,0 +1,252 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"maps"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"charm.land/log/v2"
+	"github.com/adrg/xdg"
+	"github.com/hashicorp/go-envparse"
+	"github.com/muesli/reflow/wordwrap"
+	"github.com/urfave/cli/v3"
+)
+
+var (
+	permittedKeys   = []string{"autostart"}
+	permittedValues = []string{"on", "off"}
+)
+
+func configCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "config",
+		Description: wordwrap.String("Manage Flight User Suite settings.", maxTextWidth),
+		Category:    "Configuration",
+		Commands: []*cli.Command{
+			{
+				Name:        "list",
+				Usage:       "List all Flight User Suite settings",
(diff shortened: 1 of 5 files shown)


flight-core/config_command.go
```diff
@@ -0,0 +1,252 @@
...
+	var b strings.Builder
+	for k, v := range config {
+		fmt.Fprintf(&b, "%s=%s\n", k, v)
+	}
+	return os.WriteFile(path, []byte(b.String()), 0o600)
+}
+
+func globalConfigPath() (string, error) {
+	name := filepath.Join("flight", "settings.config")
+	paths := xdg.ConfigDirs
```
jamesremuscat: I'm not sure the behaviour here is correct. 

My `$XDG_CONFIG_DIRS` is `/home/james/.config/kdedefaults:/etc/xdg/xdg-plasma:/etc/xdg` - the first entry of which is clearly user-specific - the result is that the "global" setting is being saved to somewhere in my home directory. I don't think there's any guarantee that an arbitrary entry in `$XDG_CONFIG_DIRS` is suitable for global configuration.

jamesremuscat (changes requested):
There are some quirks of how this handles global settings that I don't think are quite right.

Global settings should only be settable by root! Presumably if the first entry in `$XDG_CONFIG_DIRS` is somewhere only root-writable then a user would get an error when they attempt to set something global there, but see also below...

flight-core/config_command.go
benarmston: A better reading of the spec tells me that it is permissible to have user-specific directories in `$XDG_CONFIG_DIRS`. Sigh...

I'm generally not keen on hardcoding paths to `/etc/xdg/...`, but I think that's likely the best solution here.  I can't think of anything else.  Can you?

benarmston:
I've pushed a couple of fixes. Do you want to take a look again?

jamesremuscat (approved):
Spotted one more small thing that we could fix here, or defer, as you see fit: running `flight config get` without an argument could give a more useful error:

```
<flight> [james@jrem-laptop flight-user-suite]$ flight config get
Key 'FLIGHT_' is not known.
```

Otherwise LGTM.

PR #81, opened 2026-04-07
Augment `desktop avail` with dependency checks.

### LINKED ISSUE #76: [Flight Desktop] Indicate `doctor` results against `desktop avail` output
If the requirements for a desktop type aren't installed, it's not really "available" even if the profile is present. We should check and indicate the status of dependencies in the `avail` table output.

jamesremuscat:
This PR adds an additional column to the table output by `flight desktop avail`, showing if dependencies of each available type are available or not.

image

Resolves #76.

### DIFF
--- a/flight-desktop/type_avail.go
+++ b/flight-desktop/type_avail.go
@@ -6,6 +6,7 @@ import (
 
 	"charm.land/lipgloss/v2"
 	"charm.land/lipgloss/v2/table"
+	"charm.land/log/v2"
 	"github.com/muesli/reflow/wordwrap"
 	"github.com/urfave/cli/v3"
 )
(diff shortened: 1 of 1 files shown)


flight-desktop/type_avail.go
```diff
@@ -69,7 +71,16 @@ func typesTable(types []*Type) error {
 				hyperlink.MarginBottom(1).Hyperlink(typ.URL).Render(typ.URL),
 			),
 		)
-		t.Row(typ.ID, summary)
+
+		typ.loadDependencies()
```
benarmston: `loadDependencies()` can return an error which is being ignored here.  If it does return an error, we end up passing an empty slice of dependencies to `runDoctor`, which results in `runDoctor` claiming that all are met.

I'm not sure that's the behaviour we want, at the very least we should log the error returned by `loadDependencies`, so that running with `--log-level debug` will provide a start in understanding why there is a false positive here.

Instead of having a false positive (or false negative), we might want to add a third "Unknown" option?

benarmston (changes requested):
There's a potential error that needs to be better handled, but otherwise this is looking good.

jamesremuscat:
@benarmston I've added a commit that does as you suggested.

benarmston (approved):
Thanks for that change. LGTM.

PR #82, opened 2026-04-07
Add support for remote sessions

### LINKED ISSUE #74: [Flight Desktop] Support remote sessions
The previous incarnation of `flight-desktop` supported identifying sessions as `remote`. We currently don't, instead labelling them as `exited`. We should support remote sessions.

(I have no idea what this means)

benarmston:
It is expected that all nodes in the cluster will have access to the same home directory for the user and hence have access to the same directory where the flight desktop session data is stored.

This is useful as it allows for a user to run `flight desktop list` on a login node and get the connection details for a session started on, say, one of the compute nodes.  Sessions running on the current host are known as "local" sessions and those running on another host are known as "remote" sessions.

We only want to allow management of desktop sessions from the node on which they are running. It is not possible to kill a remote session and it is not possible to determine if a remote session should be cleaned. Attempting to kill or clean a remote session should be handled gracefully.

### DIFF
--- a/flight-desktop/session.go
+++ b/flight-desktop/session.go
@@ -44,16 +44,18 @@ var (
 	Active sessionState = "active"
 	Broken sessionState = "broken"
 	Exited sessionState = "exited"
+	Remote sessionState = "remote"
 )
 
 type Session struct {
-	Name         string
-	SessionType  string          `yaml:"session_type"`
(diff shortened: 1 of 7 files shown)


jamesremuscat (approved):
Thanks for the explanation - makes perfect sense, and this LGTM.

PR #86, opened 2026-04-08
Managed cluster plugin

### LINKED ISSUE #83: [Managed cluster] Add managed cluster plugin
The managed cluster plugin consists of the following:

* Generic documentation about managed clusters.
* A login hook that generates symlinks from the user's home directory to other locations on the cluster, e.g., from `${HOME}/scratch` to `/scratch/${USER}`.

Symlinks:

* The symlinks should not assume that `$HOME` is `/home/$USER`.
* It is OK to assume that the target exists - creating that directory is outside the scope of this issue.
* The symlinks should be read from a config file somewhere under `/opt/flight/etc/`.

Generic documentation:

* The documentation should be made available when the hook is enabled and be made unavailable when the hook is disabled.

benarmston:
This PR adds a new activation hook that creates a number of symlinks in the user's home directory to other parts of the system. E.g., `~/scratch -> /scratch/<username>`.

There is also a script to create the target directory with the correct ownership and permissions.  This script is not called directly by Flight User Suite, but can be integrated into PAM so that the directories are created when a user logs into the system.

A howto guide has been provided which documents, both the admin setup and the end result for the user.

The PAM integration ~is currently undergoing testing~ has been tested as working on my barely-a-cluster.

Resolves #83


benarmston:
@ColonelPanics the managed cluster documentation can be found on this PR.

benarmston:
@ColonelPanics the PAM integration works as long as the `/scratch` directory already exists, which is a perfectly reasonable requirement.  The howto guide should probably document that assumption though.  Do you want to include that in your adjustments to it?  Or do you want me to add that in?

ColonelPanics:
Feel free to add it, I haven't got around to this yet

managed-cluster/opt/flight/etc/user-symlinks.config
```diff
@@ -0,0 +1,10 @@
+# Symlinks created by the managed cluster hook when the Flight environment is
+# activated.
+#
+# Each line is of the format: <link_name>=<target>.  The placeholders $HOME and
```
jamesremuscat, line 4: Is this deliberately the opposite way around to `ln`'s arguments? 😛

jamesremuscat (approved):
LGTM.

managed-cluster/opt/flight/etc/user-symlinks.config
benarmston: Not *deliberately*.

benarmston:
Following discussion with @ColonelPanics there are now to howto guides for this plugin.  One is intended for end users and the other is intended for admins.

Howto guides have gained the ability to specify that they are admin only and `flight howto` respects that.

benarmston:
This probably needs another look @jamesremuscat

jamesremuscat (approved):
LGTM (merge conflicts notwithstanding); using "frontmatter" is a nice approach.

Not sure how well parsing all the howtos on demand every time will scale but it works fine for now!

benarmston:
> Not sure how well parsing all the howtos on demand every time will scale but it works fine for now!

I'm hoping that we're not going to have so many that it becomes an issue :crossed_fingers:

Issue #10, opened 2026-03-23
[Flight Core] Allow dynamic specification of tool synopsis text


jamesremuscat:
Tool synopses for help text are currently loaded from a `tool_synopsis.txt` file.

It should be possible for tools to specify this text themselves; in particular thinking of any bespoke per-cluster tooling that will get added post-install.

PR #87, opened 2026-04-08
Improved hooks and tools information

### LINKED ISSUE #26: [Flight Core] Improve hooks information
Obtaining information about hooks currently requires running multiple commands and comparing their output.  We should have a single command that can return all of the information in a more easily digested format.

The output of `flight hooks list` should be a table detailing the name, what event they're for and if enabled or not. Something like the table below but, perhaps, with more orange.

| Name | Event | Enabled |
| ---- | ----- | ------- |
| welcome | login | ✅ |
| ssh-keypair-generation | activation | ❌ |

This implies that the usage for `flight hooks list` needs to change.  Currently it takes a mandatory `<event>` argument.  This argument should no longer be mandatory or removed entirely.  There are three options:

* Remove entirely. Filtering by event is not supported. The usage becomes
  ```bash
  flight hooks list [--enabled]
  ```
* Make it optional. If present it filters the hooks to only show those trigged by `<event>`.  The usage becomes
  ```bash
  flight hooks list [event] [--enabled]
  ```
* Change it from an argument to a flag. If present it filters the hooks to only show those trigged by the given `<event>`.  The usage becomes
  ```bash
  flight hooks list [--event=<event>] [--enabled]
  ```

Of, these I think the last is the better option, but also slightly more work than the others.

---

### Non-tty outupts

If standard output is not a TTY, we ought to use a format that is more machine readable.  Something like the following, where `\t` represents a tab character.

```
welcome\tlogin\ttrue
ssh-keypair-generation\tactivation\tfalse
```

benarmston:
This PR improves the display of the `hooks list` and `tools list` commands.  They now display all hooks/tools in a table and include whether they are enabled or not.  For hooks the event is also included.

image

Filters are available to show only enabled hooks/tools or only hooks for the given event.

image

Note: This PR will outdate the instructions in FC.

Resolves #26


flight-core/main.go
```diff
@@ -225,7 +225,9 @@ func addAdminCommands(cmd *cli.Command, maxTextWidth int) {
 			Usage:     "Manage Flight User Suite hooks",
 			UsageText: fmt.Sprintf("%s hooks command [command options]", progName),
 			Description: wordwrap.String(
-				"Manage the Flight User Suite hooks.\n\nThere are two types of hooks: 'login' hooks and 'activation' hooks.  Enabled 'login' hooks are exectued when a login shell is started. Enabled 'activation' hooks are executed with the Flight environment is activated.",
+				`Manage the Flight User Suite hooks.
+
+There are two types of hooks: 'login' hooks and 'activation' hooks.  Enabled 'login' hooks are exectued when a login shell is started. Enabled 'activation' hooks are executed with the Flight environment is activated.`,
```
jamesremuscat: Old typo, might as well fix now:
```suggestion
There are two types of hooks: 'login' hooks and 'activation' hooks.  Enabled 'login' hooks are exectued when a login shell is started. Enabled 'activation' hooks are executed when the Flight environment is activated.`,
```

flight-core/tools.go
```diff
@@ -42,30 +44,71 @@ func listTools(ctx context.Context, cmd *cli.Command) error {
...
+			}
+			switch col {
+			case 0:
+				return style.Width(namecolWidth)
+			case 1:
+				return style.Width(13)
+			}
+			return style
+		})
+	t.Headers("Name", "Enabled")
```
jamesremuscat, line 82: Might be nice to also include the tool synopsis/description in this table, too. (Edit: Maybe best done as part of #10?)

jamesremuscat (changes requested):
I spotted a typo that wasn't introduced in this PR, but might as well be fixed in it. I've also made a suggestion you're free to ignore.

Otherwise this looks good - the new behaviour is exactly how I instinctively tried to use `flight tools list` previously, which suggests it's better now!

flight-core/tools.go
benarmston: Great idea, but I think its too late in the day for that now.

benarmston:
Rebased on `main` to resolve merge conflicts.  Don't see the need for another review.  Will merge once the CI has ran.

PR #95, opened 2026-04-09
Doc Tweaks

ColonelPanics:
Mainly separating out stuff that should be for admins from what end users would see/know about the tool. 

I know the placeholder file is overwritten by the README but I think it may not be worth doing that going forward?

I've placed Admin Guides for different tools within an admin guide dir in hopes that will lead to better sorting and make admin guides clearer -> do you reckon this will work when they all get enabled through? (btw symlink dirs don't work but symlinked files within do)

Also also not sure how dir sort works because if I name it 99-admin-guides it doesn't appear at end of list


benarmston:
Following GChat conversation, I've changed how we're creating the symlinks when hooks and tools are enabled.

We still want to change the order and remove some numbers, but that will likely come too late for another build.

PR #96, opened 2026-04-09
Better howto ordering and display

benarmston:
* Removes numbers from howto names
* Renders fight desktop howto above admin guides


flight-howto/main.go
```diff
@@ -189,8 +189,11 @@ func loadHowtos(dirPath string, user *user.User) ([]*Howto, error) {
 }
 
 func prettyFilename(filename string) string {
-	re := regexp.MustCompile(`^\d+-\s*`)
-	filename = re.ReplaceAllString(filename, "")
+	fmt.Printf("filename: %v\n", filename)
```
jamesremuscat:
```suggestion
```

Fairly sure we don't need this 😉

PR #97, opened 2026-04-09
Don't create symlinks for root user.

### LINKED ISSUE #94: Avoid creating user symlinks for root user
The `user-symlinks` hook should do nothing for the root user.

jamesremuscat:
Resolves #94.

Also fix detection of symlinks if the links are dangling.


benarmston (approved):
LGTM.

PR #98, opened 2026-04-09
Only show one welcome message when autostart is enabled.

jamesremuscat:
When autostart and the `welcome` hook are both enabled, we were printing two very similar welcome messages. Let's only print one.

### DIFF
--- a/flight-starter/opt/flight/usr/lib/hooks/login/welcome
+++ b/flight-starter/opt/flight/usr/lib/hooks/login/welcome
@@ -26,6 +26,21 @@
 # https://github.com/alces-flight/flight-user-suite
 #==============================================================================
 
+# Load in any settings set by 'flight config set'. Allowing per-user settings
+# to override global settings.
+if [ -e /etc/xdg/flight/settings.config ]; then
+  source /etc/xdg/flight/settings.config
+fi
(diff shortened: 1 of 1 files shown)


flight-starter/opt/flight/usr/lib/hooks/login/welcome
```diff
@@ -26,6 +26,21 @@
 # https://github.com/alces-flight/flight-user-suite
 #==============================================================================
 
+# Load in any settings set by 'flight config set'. Allowing per-user settings
```
benarmston, line 38: The only other place that we load in `settings.config` is done after we've loaded `flight.config`. As we're `source`ing the file, it's theoretically possible that values in `settings.config` could depend on those in `flight.config`.

Perhaps, that's not going to ever be an issue, but I'd feel better if we were consistent about load order, one way or the other.

benarmston (commented):
It works!  Below I've either commented with a pedantic bit of nonsense or a prescient bit of wisdom.

flight-starter/opt/flight/usr/lib/hooks/login/welcome
jamesremuscat: I've moved the new code to below the block that loads `flight.config` just to be sure.

benarmston (approved):

PR #101, opened 2026-04-09
Run go mod tidy/verify on all go subdirs

benarmston:
A somewhat random collection of improvements to the code base:

* Run `go mod tidy/download/verify` on all Go subprojects not just `flight-core`.
* Add basic specs for `flight-howto`. I've added what was easy to write an integration test for.
* Add specs for `flight-desktop`. Again, I've added what was easy to write integration tests for.
* Improve config handling for `flight-desktop`.
  * Load once per run.
  * Fail early if there is a unmarshalling error.
  * Fall back to default config if file is missing.
* Add missing README for `flight-desktop`.

Regarding tests, most of them are golden (or snapshot) tests.  The compare the actual output to a "golden file" found in `testdata/golden/`.  The files can be updated by running `go test ./... -count 1 -update`.  The `testdata` directory also contains setup data that is copied into a temp directory and used as the `FLIGHT_ROOT`.  This allows isolating the test from prior test runs and from the system state itself.

`flight-desktop` also has some non-golden tests.  One that checks that certain artefacts have been created and another that runs through a lifecycle of list -> start -> list -> show -> kill -> list (or similar).

The tests do nothing to check that the `libexec/desktop/vncserver` actually works, nor do they check that any of the `session.sh` scripts work.  That is very much left as a manual testing exercise.


jamesremuscat (approved):
Thanks for this - it's good to have some tests in place (though I'm going to have to get used to the Go style of testing) and the other changes look sensible too.

PR #102, opened 2026-04-10
Fix issue with howto removal

benarmston:
Disabling a tool or hook currently removes all howtos installed by enabling a hook or tool.  The underlying issue is that `filepath.Rel()` was returning a path with as many `../` sections as needed to get a relative path.  Resulting in howtos from other tools/hooks being erroneously determined to belong to the one being disabled.

Now we clean the howto path and check if it has a prefix matching the cleaned tool/hook's howto directory.  This works correctly, as long as the symlinks are created with absolute paths, which they are.

### DIFF
--- a/flight-core/howto_guides.go
+++ b/flight-core/howto_guides.go
@@ -5,6 +5,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"charm.land/log/v2"
 )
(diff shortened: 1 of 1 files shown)


jamesremuscat (approved):
> Disabling a tool or hook currently removes all howtos installed by enabling a hook or tool.

Oops...

This LGTM.

PR #103, opened 2026-04-10
Add `desktop logs` command.

### LINKED ISSUE #47: [Flight Desktop] Make session logs accessible
We would like to expose desktop session logs via a new `flight desktop log` command.

jamesremuscat:
image

We don't concern ourselves with pagination as it's entirely suitable for piping into `less`. Essentially, it stops people needing to figure out the session directory themselves in order to view the logs.

Resolves #47.

### DIFF
--- a/flight-desktop/session_logs.go
+++ b/flight-desktop/session_logs.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/muesli/reflow/wordwrap"
+	"github.com/urfave/cli/v3"
+)
+
+func showSessionLogsCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "logs",
+		Usage:       "Show desktop session logs",
+		Description: wordwrap.String("Display logs for a desktop session.", maxTextWidth),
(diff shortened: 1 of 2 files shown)


flight-desktop/session_logs.go
```diff
@@ -0,0 +1,50 @@
...
+				return fmt.Errorf("trying to read log file: %w", err)
+			}
+			defer log.Close()
+
+			scanner := bufio.NewScanner(log)
+
+			for scanner.Scan() {
+				line := scanner.Text()
+				fmt.Println(line)
+			}
```
benarmston, line 45: What's the reason that you've used a scanner instead of something like:

```go
			data, err := os.ReadFile(logFile)
			if err != nil {
				return fmt.Errorf("reading log file: %w", err)
			}
			fmt.Print(string(data))
```

is it just to avoid loading all of the logs in memory at once?

benarmston (approved):
LGTM. I've got a question about the use of `*bufio.Scanner`, but I'm not against it.

flight-desktop/session_logs.go
jamesremuscat: Yes, exactly that; I'm assuming the log file could potentially become quite large.

PR #108, opened 2026-04-13
Fix welcome hook and non interactive shells

### LINKED ISSUE #105: [Flight Starter] Welcome hook interferes with running commands over SSH
Stu has [found](https://chat.google.com/room/AAQAEObPY2Q/rLAgfthw7I4/rLAgfthw7I4?cls=10) some unwanted behaviour when running commands over SSH:

```
<flight> [rocky@login1 ~]$ ssh node01 '/opt/flight/bin/flight desktop list'
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
No desktop sessions found.
```

He posits:

> In a situation where I have desktop enabled and both the welcome and ssh-keypair-generation hooks enabled... Looks like it's the welcome hook, probably needs an interactive session check or something

benarmston:
This fixes an issue where `tput` error messages would be sent to STDERR when sshing to a compute node and running a command directly. E.g.,

```bash
[rocky@login1 ~]$ ssh node01 /opt/flight/bin/flight desktop list
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified
┌────────────────────┬────────┬────────────────────────────┬──────────┬────────┐
│        Name        │  Type  │     Connection string      │ Password │ State  │
├────────────────────┼────────┼────────────────────────────┼──────────┼────────┤
│ nearly-azure-xterm │ xterm  │ 10.10.0.101:5901           │ BH6W2LXT │ remote │
└────────────────────┴────────┴────────────────────────────┴──────────┴────────┘
```

An alternative fix might be to only run the hook for interactive shells. Whilst that sounds like a good solution, all of our hooks are currently executed rather than sourced. This means that the shell that is started to execute the welcome hook is never interactive.

There are some alternative possible fixes, but they all either have potentially wide ranging consequences or require considerably more work.

1. We could decide that (login?) hooks are never ran for non-interactive shells, but that would affect all (login?) hooks.
2. We could extend the hook mechanism to allow different hooks to specify whether they should be ran for non-interactive shells.
3. We could pass whether the calling shell is interactive to each hook as its first parameter.

Fixes #105

### DIFF
--- a/flight-starter/etc/profile.d/zz-flight-starter.sh
+++ b/flight-starter/etc/profile.d/zz-flight-starter.sh
@@ -28,8 +28,8 @@ if [ -z "$BASH_VERSION" ]; then
   return
 fi
 
+FLIGHT_ROOT_ORIG="${FLIGHT_ROOT}"
 export FLIGHT_ROOT=${FLIGHT_ROOT:-/opt/flight}
-FLIGHT_ROOT_ORIG=${FLIGHT_ROOT}
 
 if [ "${-#*u}" != "$-" ]; then
   set +u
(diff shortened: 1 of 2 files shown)


jamesremuscat (approved):
Makes sense to me, and solves the problem at hand.

PR #110, opened 2026-04-13
Howto guides can override default name

### LINKED ISSUE #109: [Flight howto] Support acronyms in howto filenames
When the `flight-mfa` tool is enabled, `flight howto list` shows:

```bash
$ ./opt/flight/bin/flight howto list
┌───────┬──────────────────────────────────────────────┐
│ Index │                    Title                     │
├───────┼──────────────────────────────────────────────┤
│ 1     │ About Flight User Suite                      │
│ 2     │ Flight Mfa                                   │
└───────┴──────────────────────────────────────────────┘
```

We'd like that to be "Flight MFA" not "Flight Mfa".  Similarly for any other acronyms in any other howto guides.

benarmston:
If the default name for a howto guide is not suitable, it can be overridden by specifying its name in its front matter. E.g.,

```
---
name: "Flight MFA"
---
```

Or

```
---
admin: true
name: "Admin Guides > Flight MFA"
---
```

As the second example implies, if the name is provided by front matter, it needs to provide the exact name that will be displayed in the `hwoto list` output, not just the final part.  This provides full flexibility, at the expense of a small amount of inconvenience.

Resolves #109

### DIFF
--- a/flight-howto/guides.go
+++ b/flight-howto/guides.go
@@ -4,8 +4,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"charm.land/log/v2"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


jamesremuscat (approved):
LGTM.

PR #112, opened 2026-04-13
Add Bash shell completion

benarmston:
This PR adds support for tab completion for the Bash shell.

The particulars of how commands complete can be found on #111.

The implementation is relatively straightforward, but a couple of things are worth calling out.

During the build process, bash completion configuration is generated and installed at `/etc/bash_completion.d/flight`.  This is only generated for the `flight` command, e.g., `flight-core`.

If a user types `flight <tool> <tab>`, a request for completions is sent to the `flight` command by the shell.  `flight` proxies that request to the tool by running `$FLIGHT_ROOT/usr/lib/flight-core/flight-<tool> --generate-shell-completion`.  This results in shell completion working for all of our tools as long as they are accessed via `flight <tool>`.

---

A completion script can be generated for ZSH by running:

```bash
/opt/flight/bin/flight completion zsh
```

It is left as an exercise for ZSH users to figure out where that output should be placed.

Similarly, a completion script can be generated for Fish by running:

```bash
/opt/flight/bin/flight completion fish
```

Resolves #111

PR #116, opened 2026-04-14
Improved completion inside of `flight shell`

benarmston:
Based on branch under review in #112.

Tab completion now works correctly inside of `flight shell` too.


jamesremuscat (approved):
Works well enough for me!

PR #121, opened 2026-04-15
Add `flight web start` command and hello world Web Suite service

benarmston:
A new Web Suite service has been added.  Currently it runs on a given port (default 8080) on all interfaces.  It responds to any request with "Hello, World!".

A new `web` tool has been added which allows for managing the Web Suite service.  This tool is enabled in the usual way: `flight tools enable web`.  Once enabled, it can be used to manage the Web Suite service.  It has commands to start, stop and print the status of the web suite service.

It is expected that the `web` tool will eventually be generalised into a `service` tool used to manage any of our services, and possibly merged into `flight-core`. To that end it has been kept separate from the Web Suite service itself.

Keeping the `web` tool and the Web Suite service separate also allows for them to be better focused on their usage: `web` is a CLI tool that should be nice for users to run from the CLI including good `--help` output and user-facing output/error messages. Web Suite is a service that should deal with logs and error handling in a manner suitable for services/daemons.

The `flight tools enable` command has learned `--admin-only`.  Passing this flag enables the tool for the `root` user only.  The output of `flight tools list` has been updated accordingly.  This allows ensuring that the `flight web` tool is only available for the `root` user.

Placeholder howto guides for `flight web`/Web Suite have been added.  They will need much improvement before they're ready.  But that's outside the scope of this PR.


flight-core/service.go
```diff
@@ -0,0 +1,107 @@
...
+	ID   string
+	Name string
+}
+
+func (s *Service) ExePath() string {
+	return filepath.Join(flightRoot, "usr", "libexec", s.ID, "service")
+}
+
+func (s *Service) PidfilePath() string {
+	return filepath.Join("/", "var", "run", "flight", fmt.Sprintf("%s.pid", s.ID))
```
jamesremuscat, line 25: Why not under `FLIGHT_ROOT`?

flight-core/service_start.go
```diff
@@ -0,0 +1,34 @@
...
+		Name:        "start",
+		Usage:       "Start the Flight Web Suite service",
+		Description: wordwrap.String("Starts the Flight Web Suite service, if it is not already running.", maxTextWidth),
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			// Exec the service, print where it is running and exit.
+			service := Service{ID: "web-suite", Name: "Web Suite"}
+			fmt.Printf("Starting service %s...\n", service.Name)
+			err := service.Start(ctx)
+			if err != nil {
+				return cli.Exit(fmt.Sprintf("Starting %s service failed: %s", service.Name, err), 1)
```
jamesremuscat, line 23: This isn't reporting (at least some) errors starting the process.

It's possible to run `flight web start` multiple times and be given the success message each time, even though the underlying exec returns an exit code of 1. (Running it directly gives the expected "process with PID _X_ is still running" message.)

flight-core/service_status.go
```diff
@@ -0,0 +1,59 @@
...
+				return style.Width(namecolWidth)
+			case 1:
+				return style.Width(9)
+			}
+			return style
+		}).
+		Width(termWidth)
+	t.Headers("Name", "Status", "URL")
+
+	service := Service{ID: "web-suite", Name: "Web Suite"}
```
jamesremuscat, line 52: I'm assuming this is part of what you meant by "eventually be generalised into a `service` tool" since this table seems overkill for showing the status of a single process!

flight-web/tool/main.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func main() {
+	cli.VersionPrinter = func(cmd *cli.Command) {
+		fmt.Printf("version=%s revision=%s date=%s\n", version, commit, date)
+	}
+	cmd := &cli.Command{
+		Name:                   progName,
+		Usage:                  "Manage the Flight Web Suite service",
+		Version:                version,
```
jamesremuscat: We've not included version numbers in most of our other subcommands...

jamesremuscat (commented):
Some specific thoughts below; some more general ones up here.

I'm not sure it makes any sense to allow even the _possibility_ of enabling `flight web` for non-admins. Non-admins can't use `flight web start` (since they can't, or ought not to be able to, create the pidfile); and while they _can_ run `flight web stop` without error, it doesn't actually stop the server process started by an admin - we wouldn't want it to, but it's odd that it gives no error.

In fact, having thought through this PR, I'm not sure it makes any sense to have `flight web` as something that's enableable or disableable at all. It doesn't do anything unless explicitly `start`ed (though we'll probably want autostart at some point), so all that we're doing by requiring an `enable` is adding an extra step.

So perhaps, as per our IRL conversation earlier, this really should be something built in like `flight tools` and only available to admins that way...

flight-core/service.go
benarmston: That just felt wrong for a pidfile.  I don't have a better answer than that.

flight-core/service_start.go
benarmston: `Start` will return an error if the command cannot be started.  What we ideally want is to wait for a brief while, check that its still running and somehow get it to report the URL its listening on.  I think that's a good enhancement, that is captured in the TODO below this line, but getting this branch merged, so other work can happen on top of it seems like the more urgent requirement.

flight-core/service_status.go
benarmston: That and trying to not to have too much hardcoding to web suite.  This was mostly added like this as it was quick to copy/paste/implement.  Do you need it to be changed?

flight-web/tool/main.go
benarmston: I've included it for all of the separate tools that I've written in Go, i.e., `flight-desktop`.

Given that the tools are just executable files, I would not be surprised if at some point, on some system somewhere, for some reason, a version of a tool is "upgraded" by the unsupported and undocumented process of copying just that executable into place.  Having a `--version` command available in the tools is a defensive measure to detect that and help with diagnosing any issues.

Perhaps I'm just being overly cautious here, but I can't see an issue with including the version.

flight-core/service_status.go
jamesremuscat: Fine as is.

flight-web/tool/main.go
jamesremuscat: Your logic is sound; in which case the inconsistency is that `howto` doesn't have its own version number (which clearly isn't to be fixed here).

jamesremuscat:
> We don't want [Web Suite's howto] to be available if the site are not running the web suite.

FWIW I'd be reasonably happy for it to be available, so long as it makes it clear that it's something their admins need to turn on. (Might be a nice way for the admins to discover their users want the feature?)

flight-web/tool/main.go
jamesremuscat: (I don't know where "most" came from in my original comment, maybe I was looking in the wrong places...)

benarmston:
This is ready for another look @jamesremuscat

flight-core/main.go
```diff
@@ -210,6 +210,12 @@ func addAdminCommands(cmd *cli.Command, maxTextWidth int) {
 						),
 						maxTextWidth,
 					),
+					Flags: []cli.Flag{
+						&cli.BoolFlag{
+							Name:  "admin-only",
+							Usage: "enable the tool only for admin users",
+						},
+					},
```
jamesremuscat: This flag no longer has any effect.

```suggestion
```

jamesremuscat (changes requested):
One more small piece of tidy-up...

flight-core/main.go
jamesremuscat: (I'm not sure why GitHub is rendering the tabs twice as wide in the suggestion as in the diff...)

benarmston: That's gone now.

jamesremuscat (approved):

Issue #118, opened 2026-04-14
[Flight Web] Themed home page


jamesremuscat:
As per [design](https://www.figma.com/design/hxJJ4yBksw4dkThQevlihb/Web-suite?node-id=1125-55&t=QrXgqijC6c5krazx-4)...

<img width="3700" height="6116" alt="Image" src="https://github.com/user-attachments/assets/fd1f3dec-19d4-4d02-a3a3-299b37ddf6c5" />

Some aspects of this aren't relevant (e.g. config packs) and others will need tweaking. The internal/tool links need not be functional (or filtered) at this stage; the purpose of this issue is to get the design framework for the FWS in place.

PR #123, opened 2026-04-17
Themed home page

benarmston:
This PR includes a partial implementation of the themed home page by @megan-godfrey; some work on easing setup of a development environment; partial work on integrating into the production build by @benarmston; and some finishing off of that by @jamesremuscat.

### Themed home page

image

Initial work on the themed home page.

* List of tools is hardcoded.
* The tool cards don't link anywhere.

I (James) am not going to try and complete the design in this PR; there's enough there that we can start to bolt on functionality.

### Development process

* Download tailwindcss CLI, by running
  ```bash
  cd flight-web-suite
  ./bin/tailwind/download-tailwindcss.sh
  ```
* Download air - Live reload for Go apps
  ```bash
  go install github.com/air-verse/air@latest
  ```
(Ensure that the Go bin directory e.g. `~/go/bin` is on your `$PATH`)

* Launch tailwindcss and webserver both with live-reloading
  ```bash
  cd flight-web-suite
  air
  ```
* Edit files --- tailwind will recompile automatically and Go server will be rebuilt automatically.

### Production build

The server and all of its assets are built into `flight-web-suite/dist` and will be included in the tarball.  When running in production, the server looks for its assets and views directories under `$FLIGHT_ROOT/var/web-suite` unless it can find them in the current working directory (as it can when running under `air`).

Refs #118


jamesremuscat:
@benarmston I think there's enough here to warrant merging in this state. Do you agree?

benarmston:
> @benarmston I think there's enough here to warrant merging in this state. Do you agree?

Yes, I do.

Issue #120, opened 2026-04-14
[Flight Web] User login frontend


jamesremuscat:
Let's let cluster users log in using the mechanisms implemented in #119.

<img width="1344" height="1010" alt="Image" src="https://github.com/user-attachments/assets/fccecf6c-3916-44ce-982b-1543ca37287b" />

jamesremuscat:
Implemented in #124.

PR #124, opened 2026-04-17
User session management

benarmston:
This PR adds user session backend and frontend, along with the beginnings of a test suite.

[Screencast from 2026-04-17 18-17-01.webm](https://github.com/user-attachments/assets/9a5cf502-8ac9-4abf-9748-cb7d1de3d2dc)

The login form is similar to the image in #120. However, it was easiest to copy the sign in form from Concertim SSO and change some styling, so I did that instead of a faithful implementation.

The backend delegates authentication to a Python script which uses the systems PAM library and the PAM login service.  Using a separate executable, avoids linking the web suite service to a pam library via CGO.  CGO would significantly reduce the portability of the binary.  The Python script does not have a compile time step, so it should be portable to any system with Python3, the `python-pam` library and PAM.

The beginnings of helpers and layouts have been added.  Some of the helpers return an error whilst others log any errors and don't return them.  We probably want to decide on a preferred approach.

I've added views with the extension `.gohtml` as this is what my Nvim setup is already configured for and is a common extension for Go HTML templates. We should standardise on an extension for our templates. I can change my Nvim setup easily so that should not be a determining factor.  However, configuring it to treat `.html` files as HTML for some projects and Go HTML templates for this project is something I'd prefer not to do.

The tests I've added can be split into two categories: "functional" tests that make assertions against the response to a single request and "integration" tests that make multiple requests and assert against the responses to any or all of them.  In both cases, I've added reasonably small DSLs in the `internal/testutil` package to make writing tests more pleasant and more importantly to provide decent error messages when something fails.

The Makefile has also seen some improvements to aid in setup when running tests, the dev server and the packaged "production" server.  In particular, it avoids me having to remember to set `FLIGHT_ROOT` correctly.

## Discussion points

* Should helpers return `error` or log and swallow?
* Should views be `.gohtml` files or something else?

## Future enhancements:

* Determine how configuration is going to work and make the session secret and PAM service configurable.


flight-web-suite/middleware.go
```diff
@@ -0,0 +1,35 @@
...
+func MethodOverrideMiddleware() echo.MiddlewareFunc {
+	config := middleware.MethodOverrideConfig{
+		Getter: middleware.MethodFromForm("_method"),
+	}
+	return middleware.MethodOverrideWithConfig(config)
+}
+
+func NewSessionMiddleware() echo.MiddlewareFunc {
+	// TODO: Use a secret secret.
+	sessionStore := sessions.NewCookieStore([]byte("totally-not-a-secret"))
```
benarmston, line 20: This needs some thought on how we're going to set this to a different value for each installation. We might want an automated solution that runs when flight web start is first ran, or documentation telling the admin to set it somehow (perhaps `flight config set web.session.secret ${random_string}`) or something else.

Which ever option we take, I think this is for a separate PR.

jamesremuscat: The first option sounds preferable to me - I'll open an issue to track it.

flight-web-suite/main.go
```diff
@@ -74,21 +79,35 @@ func main() {
...
+func newApp() *echo.Echo {
 	e := echo.New()
+	e.Pre(MethodOverrideMiddleware())
 	e.Use(middleware.RequestLogger())
-	e.Renderer = &echo.TemplateRenderer{
-		Template: template.Must(template.ParseGlob(getDirectory("views") + "/*.html")),
-	}
+	e.Use(NewSessionMiddleware())
+
+	// TODO: Standardise on extension for Go html/template files.
```
jamesremuscat: Happy to standardise on `.gohtml`; I'll raise an issue to track that separately.

flight-web-suite/helpers.go
```diff
@@ -0,0 +1,90 @@
...
+	return nil
+}
+
+func SaveSession(c *echo.Context, sess *sessions.Session) {
+	if err := sess.Save(c.Request(), c.Response()); err != nil {
+		c.Logger().Warn("Error saving session", "err", err)
+	}
+}
+
+func LoggedIn(c *echo.Context) bool {
```
jamesremuscat: Pedantic but is `IsLoggedIn` a better name?

flight-web-suite/helpers.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return false
+	}
+	if username, ok := sess.Values["username"].(string); ok && username != "" {
+		return true
+	}
+	return false
+}
+
+func CurrentUser(c *echo.Context) string {
```
jamesremuscat: Again pedantry: would `CurrentUserName` be a better name? It makes it clear it's not some sort of `User` record (yet?).

.github/workflows/build.yml
jamesremuscat: It'd be a shame not to run all those nice new tests you've written!

```suggestion
        subdir: ["flight-core", "flight-desktop", "flight-howto", "flight-web-suite"]
```

jamesremuscat (changes requested):
I've made some comments and suggestions - some are pedantic, but we should probably run the new tests in CI 😉 

Also: this works so fast compared to Flight Center / Concertim SSO it feels like the authentication is faked. Might need to add a `sleep` somewhere!

We should document (somewhere) our dependency on `python` and `python-pam` and its aliases - eventually this will end up in the `web doctor` command which I need to raise an issue to track! (For the record, the Ubuntu package is called `python3-pampy` which installs a package called `pam`, and **not** the obvious `python3-pam`, which installs `pypam`. Clearly.)

flight-web-suite/main.go
benarmston: I've done it on this branch.

benarmston:
> We should document (somewhere) our dependency on `python` and `python-pam` and its aliases

I've made some very placeholdery/todoish additions to the web suite admin guide. It still needs writing correctly, but it shouldn't get forgotten.

benarmston:
Do you want to take another look at this @jamesremuscat?

jamesremuscat (approved):
LGTM!

PR #126, opened 2026-04-20
Web suite configuration

### LINKED ISSUE #129: [Flight Web] Generate web session secret on first run if needed
#124 introduces sessions to FWS, for which we need a server-side secret.

We should not use `flight config` to store this secret, since its global configuration file is necessarily publicly readable! Perhaps a file in `/opt/flight/etc/flight-web`?

If there is no secret configured, we should generate one the first time `flight web start` is run.

gorilla's docs say:

> Ensure your key is sufficiently random - i.e. use Go's crypto/rand or securecookie.GenerateRandomKey(32) and persist the result.

benarmston:
This PR adds support for configuring web suite and correctly handling its session secret.

Web suite now supports a config file at `${FLIGHT_ROOT}/etc/web-suite.yml`. It allows configuring the listen port, authenticator PAM service, and authenticator timeout.

A new environment variable, `FLIGHT_STATE_ROOT`, has been introduced, defaulting to `${FLIGHT_ROOT}/var/lib`. It is included in `/etc/xdg/flight.config`, and shared environment initialisation has been added so Flight tools can resolve `FLIGHT_ROOT` and `FLIGHT_STATE_ROOT` consistently, including support for variable expansion in `flight.config`.

When web suite starts, it loads its session secret from `${FLIGHT_STATE_ROOT}/web-suite/session-secret`. If that file does not exist, web suite generates a new secret and writes it there. This gives the session secret a single runtime location and avoids persisting generated state into tracked configuration files.

The dev/test workflow has also been updated to provide a sensible `FLIGHT_STATE_ROOT`, and web-suite tests now use temporary state directories so local runs do not dirty files in the repo or packaged tree.

As part of this work, `flight-core/pkg` has been refactored into smaller packages with narrower responsibilities. In particular, CLI/TUI code has been separated from general helpers so non-interactive code such as web suite can safely reuse shared functionality like pidfile and environment helpers without importing terminal/UI-related side effects.

Resolves #129
Refs #128


jamesremuscat:
Not sure why yet but I'm getting an HTTP 500 error when trying to log in to FWS now (having done a `make distclean dist` at the top level).

benarmston:
> Not sure why yet but I'm getting an HTTP 500 error when trying to log in to FWS now (having done a `make distclean dist` at the top level).

Has the session secret changed?  Does deleting the session cookie fix this?

jamesremuscat:
> Has the session secret changed? Does deleting the session cookie fix this?

Good guess! That sorted it. (I'd assumed it was something wrong with my environment setup and was fiddling around with that...)

We should probably do something a bit nicer than 500 if someone's tried to forge a session cookie!

jamesremuscat (approved):
Thanks for that sessions change - this LGTM.

benarmston:
Rebased to resolve a merge conflict.  I'll merge once CI passes.

PR #134, opened 2026-04-22
Only show enabled tools

jamesremuscat:
This PR restricts the display of tools to those which are enabled; it also requires the user be logged in to see them.

It does so by duplicating functionality from `flight-core`; ideally this would be shared functionality, which should be easier once #126 is merged (so I've deferred it from here). An aborted attempt at doing so is in 00252fd which this PR also reverts.

We currently rely on the fact that all admin-only functionality for `flight` is built-in and there are no admin-only tools - we just check that the tool is enabled, not that the specific user has access to it. The latter will likely be easier to do once other user-specific functionality is in place, should we ever need it.

Of low-to-moderate interest may be the `setFlightRootForTest` method used in `home_functional_test.go` to set up a faked Flight environment that has both `howto` and `desktop` "enabled". EDIT: That also appears, in a modern form, in #126... sigh...

### DIFF
--- a/flight-web-suite/home_functional_test.go
+++ b/flight-web-suite/home_functional_test.go
@@ -16,6 +16,7 @@ func TestHomepageAnonymous(t *testing.T) {
 }
 
 func TestHomepageAuthenticated(t *testing.T) {
+	setFlightRootForTest(t, "./testdata/flight_root") // Fake FLIGHT_ROOT with dummy tools enabled
 	_, body := testutil.RenderPage(t, newApp(), http.MethodGet, "/", nil, http.StatusOK, testutil.WithSessionCookie("ben"))
 
 	assertAuthenticated(t, body, "ben")
@@ -70,4 +71,14 @@ func assertAuthenticated(t *testing.T, body, username string) {
 		testutil.HasText("Logout"),
 	)
 	testutil.AssertNoSelection(t, body, `[data-testid="sign-in-link"]`)
+}
+
+func setFlightRootForTest(t *testing.T, path string) {
+	t.Helper()
+
+	origPath := flightRoot
+	flightRoot = path
+	t.Cleanup(func() {
+		flightRoot = origPath
+	})
 }
--- a/flight-web-suite/home_functional_test.go
+++ b/flight-web-suite/home_functional_test.go
@@ -11,8 +11,8 @@ func TestHomepageAnonymous(t *testing.T) {
 	_, body := testutil.RenderPage(t, newApp(), http.MethodGet, "/", nil, http.StatusOK)
 
 	assertNotAuthenticated(t, body)
-	assertToolCardPresentHTML(t, body, "Flight Desktop", "/assets/images/desktop.png", "Access interactive desktop sessions")
-	assertToolCardPresentHTML(t, body, "Flight Howto", "/assets/images/howto.png", "Learn about the Flight User Suite and using your cluster")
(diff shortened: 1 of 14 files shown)


flight-web-suite/views/home.gohtml
```diff
@@ -4,23 +4,31 @@
...
-	  flex flex-col items-center gap-y-1 justify-center
-	  border border-alces-blue-light
-	  w-70 p-9 px-8
-	  hover:shadow hover:border-alces-blue transition-colors duration-200 cursor-pointer group
-	" data-testid="tool-card--{{ .Name }}">
-        <div class="size-25 mb-2">
-            <img src="{{ .IconPath }}" alt="{{ .Name }} logo" class="h-full">
+
+{{ if .CurrentUserName }}
+    <p class="mb-8">Select a tool to access your cluster from within your browser.</p>
```
benarmston: I'm being very picky, but this doesn't handle the case where there are no tools enabled or an error in listing the enabled tools.  Perhaps we should?

flight-web-suite/home_functional_test.go
```diff
@@ -11,11 +11,12 @@ func TestHomepageAnonymous(t *testing.T) {
...
 
 	assertNotAuthenticated(t, body)
-	assertToolCardPresentHTML(t, body, "Flight Desktop", "/assets/images/desktop.png", "Access interactive desktop sessions")
-	assertToolCardPresentHTML(t, body, "Flight Howto", "/assets/images/howto.png", "Learn about the Flight User Suite and using your cluster")
+	assertToolCardAbsentHTML(t, body, "Flight Desktop")
+	assertToolCardAbsentHTML(t, body, "Flight Howto")
 }
 
 func TestHomepageAuthenticated(t *testing.T) {
+	setFlightRootForTest(t, "./testdata/flight_root") // Fake FLIGHT_ROOT with dummy tools enabled
```
benarmston, line 19: Nice!

flight-web-suite/main.go
```diff
@@ -110,21 +103,20 @@ func newApp() *echo.Echo {
...
-				IconPath:    "/assets/images/howto.png",
-			},
-		},
+		"Tools":   getToolsList(),
+	}
+}
+
+func getToolsList() []Tool {
+	flightTools, _ := getTools(true)
+	availableTools := make([]Tool, 0, 2)
```
benarmston: If we're setting the capacity of the slice, we ought to determine that from the length of the `flightTools` slice.

```suggestion
	availableTools := make([]Tool, 0, len(flightTools))
```

flight-web-suite/main.go
```diff
@@ -110,21 +103,20 @@ func newApp() *echo.Echo {
...
-				URL:         "/howto",
-				IconPath:    "/assets/images/howto.png",
-			},
-		},
+		"Tools":   getToolsList(),
+	}
+}
+
+func getToolsList() []Tool {
+	flightTools, _ := getTools(true)
```
benarmston: I agree that if there's an error getting the tools there are zero tools to list.  However, it'd be really useful to log the error if there is one.

benarmston (changes requested):
It looks good and works well. I have a few suggestions I'd like you to look at please.

jamesremuscat:
@benarmston I've addressed your comments. Would you like another look?

benarmston (approved):
Thanks for those changes. LGTM.

PR #136, opened 2026-04-22
Web suite can terminate local active desktops sessions

### LINKED ISSUE #132: [Flight Web - Desktop] Terminate a running VNC session
The user should be able to terminate a _local_ running VNC session.

[Design reference](https://www.figma.com/design/hxJJ4yBksw4dkThQevlihb/Web-suite?node-id=136-93&t=UMPp9CqqhiakHlnz-4)

benarmston:
This PR adds support for terminating local active desktop sessions.

image

Active sessions have a "Terminate" entry.

image

Clicking on it reloads the page and displays a success (or failure) flash.

image

Termination of remote desktop sessions is not yet implemented.

image

Cleaning up of stale (exited or broken) sessions is likewise not-yet implemented.

image

### Implementation

The `desktop kill` command has learned a new `--format FORMAT` flag.  This is similar to the flag for the `list` command.  It has two valid values `json` and `pretty`. The default is `pretty` and provides the existing behaviour. The `json` format is intended for consumption by other parts of FUS.

Executing `desktop kill` is largely similar to executing `desktop list`. Some minor refactoring has been done to provide a common way to run the `flight-desktop` executable. 

Finally, a change has been made to the format of permitted desktop session names.  Previously, `--foo` was a valid desktop name.  Allowing desktop names to begin with `-` (or `--`) means that care needs to be taken to avoid `flight-desktop` parsing them as arguments.  Web suite takes care to avoid that, but I've also changed the validation on session names to prevent names from starting with `-`.  I can't see this causing any issues and makes the lives of CLI users of `flight-desktop` easier.

### Known issues and missing features

The implementation of the dropdown menu leaves a lot to be desired. Its main advantage is that it was simple to implement and doesn't require javascript.  Eventually, we'll want to have javascript included, but that is outside the scope of this PR.

Termination of remote desktop sessions is not yet implemented.

Cleaning up of exited or broken sessions is not yet implemented.

### DIFF
--- a/flight-desktop/main.go
+++ b/flight-desktop/main.go
@@ -44,6 +44,18 @@ func init() {
 	maxTextWidth = min(termWidth, 80)
 }
 
+var formatFlag = &cli.StringFlag{
+	Name:  "format",
+	Value: "pretty",
+	Usage: "use specified `FORMAT` for the output (pretty, json).",
+	Validator: func(format string) error {
+		if format != "pretty" && format != "json" {
+			return fmt.Errorf("%s is not a known format (pretty, json)", format)
+		}
+		return nil
+	},
+}
+
 func main() {
 	// Load config here rather than `init` to prevent errors from breaking
 	// tests.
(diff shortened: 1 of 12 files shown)


jamesremuscat (approved):
This LGTM.

PR #138, opened 2026-04-23
Flight Web can launch local desktop sessions

benarmston:
This PR adds support to web suite to launch local desktop sessions.

[Screencast from 2026-04-24 13-21-42.webm](https://github.com/user-attachments/assets/7e57a8c3-551c-4b56-8c2b-167cd5327eff)

It does that by teaching `flight desktop avail` to return machine-readable JSON, then using that API to build a new launch form.  The `flight desktop start` command has also learned to return machine-readable JSON and is used when processing the submitted form.

The JSON output schema for `flight start` is more involved than it is for other commands as it has a number of failure paths, some of which are exercised by user provided input and others that are not.  The more complicated error schema allows the web UI to present structured, user-appropriate failures instead of raw CLI messages. The error format is modeled loosely on JSON:API error objects, so should be reasonably familiar.

The web-side desktop command integration has been extracted into a dedicated package.

### Limitations

- Available geometries are a hardcoded list.
- The web flow currently calls `desktop avail` twice: once to render the form and again to validate submitted desktop type input.  This is fast enough that it might not be a problem.
- The most likely user-correctable launch failure is still an invalid name; dependency-related failures are only reported after submission.

### Possible enhancements

- Improve handling for missing global desktop dependencies, possibly by disabling the form.
- Improve handling for missing type-specific dependencies, possibly by disabling or hiding affected desktop types.
- Derive supported geometries dynamically, for example via xrandr.
- Add client-side validation for session-name format.

### DIFF
--- a/flight-web-suite/desktop/build_command.go
+++ b/flight-web-suite/desktop/build_command.go
@@ -1,0 +1,91 @@
+// Package desktop provides functions and types for running the
+// `flight-desktop` CLI and parsing its output.
+package desktop
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
+	"slices"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/concertim/flight-user-suite/flight/configenv"
+)
+
+func desktopToolPath(env configenv.Env) string {
+	return filepath.Join(env.FlightRoot, "usr", "lib", "flight-core", "flight-desktop")
+}
+
+func buildDesktopCommand(ctx context.Context, env configenv.Env, username string, args ...string) (*exec.Cmd, error) {
+	userInfo, err := user.Lookup(username)
+	if err != nil {
+		return nil, fmt.Errorf("looking up user %q: %w", username, err)
+	}
+
+	uid, err := strconv.Atoi(userInfo.Uid)
+	if err != nil {
+		return nil, fmt.Errorf("parsing uid for user %q: %w", userInfo.Username, err)
+	}
+	gid, err := strconv.Atoi(userInfo.Gid)
+	if err != nil {
+		return nil, fmt.Errorf("parsing gid for user %q: %w", userInfo.Username, err)
+	}
+
+	groupIDs, err := userInfo.GroupIds()
+	if err != nil {
+		return nil, fmt.Errorf("looking up groups for user %q: %w", userInfo.Username, err)
+	}
+	groups := make([]uint32, 0, len(groupIDs))
+	for _, groupID := range groupIDs {
+		parsed, err := strconv.Atoi(groupID)
+		if err != nil {
+			return nil, fmt.Errorf("parsing group id %q for user %q: %w", groupID, userInfo.Username, err)
+		}
+		groups = append(groups, uint32(parsed))
+	}
+
+	cmd := exec.CommandContext(ctx, desktopToolPath(env), args...)
+	cmd.Dir = userInfo.HomeDir
+	cmd.Env = commandEnv(userInfo)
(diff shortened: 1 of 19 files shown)


flight-web-suite/desktop/build_command.go
```diff
@@ -0,0 +1,91 @@
...
+	if os.Geteuid() == 0 {
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Credential: &syscall.Credential{
+				Uid:    uint32(uid),
+				Gid:    uint32(gid),
+				Groups: groups,
+			},
+		}
+	} else if os.Geteuid() != uid {
+		return nil, fmt.Errorf("cannot run desktop listing as %q without root privileges", userInfo.Username)
```
jamesremuscat:
```suggestion
		return nil, fmt.Errorf("cannot run desktop command as %q without root privileges", userInfo.Username)
```

flight-web-suite/desktop/launch.go
```diff
@@ -0,0 +1,118 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/concertim/flight-user-suite/flight/configenv"
+)
+
+type Type struct {
```
jamesremuscat, line 14: I have a slight ick at calling the type `Type` but not sure it's strong enough to ask you to change it to `DesktopType`...

jamesremuscat (changes requested):
This works well and looks OK - I've made a couple of suggestions below but nothing too significant.

I'll raise a bunch of issues for some of your suggested enhancements (and some of my own).

jamesremuscat:
> Derive supported geometries dynamically, for example via xrandr.

I have no problem (via `flight desktop start`) starting sessions with very non-standard geometries (e.g. 1450x400); perhaps there are limits on the maximum size we need to worry about though?

benarmston:
> > Derive supported geometries dynamically, for example via xrandr.
> 
> I have no problem (via `flight desktop start`) starting sessions with very non-standard geometries (e.g. 1450x400); perhaps there are limits on the maximum size we need to worry about though?

I'm not sure I understand what you're getting at here?

flight-web-suite/desktop/launch.go
benarmston: I'll include it in an upcoming PR. I don't want to switch branches right now.

jamesremuscat:
> I'm not sure I understand what you're getting at here?

What I'm getting at is that I'm not sure I understand what "supported geometries" are, as I can happily start VNC sessions with random resolutions not mentioned by `xrandr`.

benarmston:
> What I'm getting at is that I'm not sure I understand what "supported geometries" are, as I can happily start VNC sessions with random resolutions not mentioned by `xrandr`.

That works for me too.  Every tool for setting resolution that I'm aware of provides a dropdown instead of supporting arbitrary resolutions.  So I'm concerned that it might not be widely supported.

I'd also, suggest that asking the user to select an option is a better UX than asking them to remember sensible resolutions.  Perhaps "Large", "Small", "Medium" might be an even better UX?

jamesremuscat:
> I'd also, suggest that asking the user to select an option is a better UX than asking them to remember sensible resolutions. Perhaps "Large", "Small", "Medium" might be an even better UX?

Maybe, but I can also imagine user requirements of "I need to use piece of software _X_ that states it needs a desktop resolution of at least 1024x768" so I don't think we should abstract away the numbers entirely.

Perhaps:
- Small (1024x768)
- Medium (1280x720 or whatever the closest VESA size is?)
- Large (1920x1080)

and have custom resolution as a future enhancement?

PR #149, opened 2026-04-28
Improved template rendering

benarmston:
This PR improves template rendering and adds a JS pipeline.  The motivation for this is to support connections to desktop sessions, which requires the NoVNC javascript library and for a different page layout for the session page.

To add JavaScript to the build, write a new JS file in `assets/javascripts` and `import` it in `assets/javascripts/application.js`.

Template rendering has been reworked to add jinja2-inspired template inheritance and custom layouts.

* Templates have split into "layouts", "partials" and "pages".
* On startup, all layout and partial templates are parsed.
* Layouts define the layout for the page along with a "content" block that pages redefine.
* All pages define "content" that is plugged into the layout.  As all pages define this, they necessarily conflict, so only a single page can be added to the template set when rendering.
* On each request, the base set of templates is cloned and the single page being rendered is added. This ensures when the layout is rendered, there is a single redefinition of "content" that is used in the layout.
* If the `data` argument contains a "layout" entry that layout is used to render the template. The "application" layout is used by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@
 flight-user-suite*.tar.gz
 flight-web-suite/tmp/*
 flight-web-suite/static/*
+flight-web-suite/node_modules
 flight-web-suite/bin/tailwind/tailwindcss
 *.DS_Store*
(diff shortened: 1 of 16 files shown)


megan-godfrey (approved):
LGTM!

Issue #137, opened 2026-04-23
[Flight Web - Desktop] Connect to running VNC session


jamesremuscat:
Users should be able to connect to and use running VNC sessions.

In order to connect to a running VNC session, we'll need to:
- Start a `websockify` process to translate between the VNC server's TCP port and an HTTP/WebSocket port
  - `websockify` can proxy to a process on another host, so this should work equally well for remote sessions
- Proxy from the FWS server to the Websockify port (see PoC in #125)
  - Although the VNC session itself will have a password, we might want to consider also restricting access to the proxied session to the correct user?
- Host a JavaScript VNC client such as [NoVNC](https://github.com/novnc/noVNC/) on a page in FWS and have it connect to Websockify

benarmston:
From figma (https://www.figma.com/design/hxJJ4yBksw4dkThQevlihb/Web-suite?node-id=0-1&p=f&t=5Y7sEjS47CGusPaF-0)

<img width="1498" height="903" alt="Image" src="https://github.com/user-attachments/assets/6b1f604e-f624-41f4-816a-72e83bf52483" />

jamesremuscat:
Basics implemented in #151 - subsequent enhancements raised in a number of separate issues.

PR #151, opened 2026-04-29
Web suite can connect to desktop sessions

benarmston:
This PR supports connecting to a desktop session.

* Start a websockify process when a desktop session starts.
* `desktop show <session>` has learned a `--format` flag. This can be used to return a machine-readable representation of a desktop session, including the host and port of its associated websockify process.
* Add a "View" button to the desktop cards which navigates to the desktop session show page
* Add a desktop show page which uses NoVNC to connect to `ws://<fus_host>:<fus_port>/websockify?host=<websocket_host>&port=<websocket_port>`.
* Add new route to FUS which proxies requests arriving at `/websockify` to a server at `http://<host>:<port>` where host and port are taken from the GET parameters.

TODO:

* Add better user-facing status indications: "connecting", "disconnected", etc..
* Automatically reconnect if connection is dropped.
* Support full screen and/or zen mode.
* Improve layout of show page. In particular, use more of the available width.
* Support termination within show page.
* Support resizing within show page.
* Maybe support renaming?
* Add support for grabbing screenshots and displaying on cards and whilst connecting to/disconnected from session.
* Add support for starting websockify after the fact. If a session has been started before `websockify` has been installed, the VNC session will be running without websockify. We'd like to be able to start websockify by running, say, `flight desktop webify <sessionName>`.

### DIFF
--- a/flight-web-suite/assets/javascript/novnc-wrapper.js
+++ b/flight-web-suite/assets/javascript/novnc-wrapper.js
@@ -1,0 +1,185 @@
+import RFB from '@novnc/novnc'
+
+function noop() {}
+
+function createConnection({
+  url,
+  domEl,
+  onClipboard=noop,
+  onDisconnect=noop,
+  onConnect=noop,
+  onPasswordPrompt=noop,
+  password,
+  viewOnly,
+}) {
+  let rfb = null;
+  try {
+    console.log('connecting to', url);
+    rfb = new RFB(
+      domEl,
+      url,
+      password && {credentials: {password}}
+    );
+    rfb.addEventListener('connect', onConnect);
+    rfb.addEventListener('disconnect', onDisconnect);
+    rfb.addEventListener('credentialsrequired', onPasswordPrompt);
+    rfb.addEventListener('clipboard', onClipboard);
+    // TODO: Add this back. Requires set height and width for domEl (not just 100%).
+    // rfb.scaleViewport = true;
+    rfb.resizeSession = false;
+    rfb.viewOnly = viewOnly;
+  } catch (err) {
+    console.error(`Unable to create RFB client: ${err}`)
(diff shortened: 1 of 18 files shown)


flight-web-suite/desktop_sessions.go
```diff
@@ -50,6 +50,28 @@ func indexDesktopSessionsHandler(c *echo.Context) error {
...
+	if err := requireDesktopToolEnabled(); err != nil {
+		return err
+	}
+
+	session, err := desktop.ShowCommand(c.Request().Context(), env, CurrentUserName(c), c.Param("sessionName"))
+	if err != nil {
+		return err
+	}
+
+	// TODO: Redirect and flash if session cannot be found or is terminated etc..
```
jamesremuscat, line 66: We'll also want to handle a session terminating while being viewed (from the JS side obviously - your comment was just a convenient point to attach this comment to!)

flight-web-suite/assets/javascript/novnc-wrapper.js
```diff
@@ -0,0 +1,184 @@
...
+}
+
+function initNoVnc() {
+  const domEl = document.getElementById("novnc-wrapper")
+  if (domEl == null) {
+    return
+  }
+  console.log("initialising NoVNC")
+
+  // Read parameters specified in the URL query string
```
jamesremuscat: This comment confused me until I realised it meant:

```suggestion
  // Read parameters to be specified in the URL query string passed to NoVNC
```

jamesremuscat (approved):
This works well - as you say, still a lot of polish to be applied and I've made a handful of suggestions to add to the list - I think we can track them as separate issues, which will let us prioritise. I'm sure Stu will have many of his own to add too!

benarmston:
Refs #137

PR #152, opened 2026-04-29
Fix detection of light terminal backgrounds

benarmston:
Upgrading lipgloss appears to have broken this. This fix is simple.


jamesremuscat (approved):
LGTM. And actually works in my terminal now!

PR #160, opened 2026-04-30
Don't show disabled tools in the help output

benarmston:
Fix issue with disabled tools appearing in `--help` output.


benarmston:
Tiny change. Merged without review.

PR #165, opened 2026-04-30
Improve handling of missing dependencies

### LINKED ISSUE #147: [Flight Web - Desktop] Improve handling of missing desktop dependencies
If a desktop type has missing dependencies, we should disable it in the web UI. (But leave it visible and with an explanatory message - or else users won't know it's a thing they could have, and won't ask their admins to fix it).

I'm not sure if there's value in distinguishing between global and type-specific dependencies. In either scenario the resolution is "ask your admin to fix it"! If we do decide to distinguish the two then let's follow the same approach (showing the Flight Desktop button but disabling with a message).

megan-godfrey:
Warning messages are shown on the launch form if any or all of the desktop types are unavailable

Screenshot 2026-04-30 at 17 18 25
Screenshot 2026-04-30 at 17 22 14

All desktop types are disabled and an error shown if the core dependencies for Flight Desktop are not met
Screenshot from 2026-05-06 12-30-14

Resolves #147

### DIFF
--- a/flight-desktop/doctor.go
+++ b/flight-desktop/doctor.go
@@ -25,9 +25,9 @@ func doctorCommand() *cli.Command {
 		Flags:       []cli.Flag{formatFlag},
 		Action: func(ctx context.Context, cmd *cli.Command) error {
 			if cmd.String("format") == "json" {
-            	report := buildDoctorReport(ctx)
-            	return writeChecksJSON(report)
-            }
+				report := buildDoctorReport(ctx)
+				return writeChecksJSON(report)
+			}
 			greenText := lipgloss.NewStyle().Foreground(lipgloss.Green)
 			redText := lipgloss.NewStyle().Foreground(lipgloss.Red)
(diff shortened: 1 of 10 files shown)


megan-godfrey:
@jamesremuscat does this fit with what you had in mind for this issue?

jamesremuscat (approved):
> @jamesremuscat does this fit with what you had in mind for this issue?

Spot on!

jamesremuscat (changes requested):
I prematurely clicked "approve"...

We do still need to consider global dependencies (i.e. those of `flight desktop` rather than the specific desktop types). If there's a missing global dependency, we should ~either~ disable all types, ~or disable Desktop altogether.~ [Stu makes the good point that remote desktop sessions could be running without deps being met on the FWS host node, so don't disable Desktop entirely!]

jamesremuscat (approved):
Thanks for those changes - this now LGTM!

PR #167, opened 2026-05-01
Better FWS launch feedback

### LINKED ISSUE #162: [Flight Web] Improve `flight web start` feedback
At present the output of `flight web start` can be misleading:
- It claims the server is started and listening, even if the server didn't launch correctly
- It hard-codes the interface and port regardless of where the server is actually listening

Running `flight web start`:
- should check that the server has started successfully before claiming it has
- use the actual interface and port in use by the server
- ideally, report errors if any are returned from the server process

jamesremuscat:
This PR makes improvements to launching FWS with `flight web start`.

- We check for an existing web process via the expected pidfile, and error out early if it's already running.
- Starting and stopping both use a spinner to show progress.
- Once successfully launched, the actual address and port being bound by the server are returned to the starting process and displayed to the user.
- If launching the server returns errors, these are caught and displayed to the user (rather than silently swallowed).

As a side-effect of these changes, the server will log an error if started standalone (i.e. `/opt/flight/usr/libexec/web-suite/service` is invoked directly, or via `air` in development) as it has no back-channel file descriptor to report the server start to. This is harmless and does not prevent the server from starting.

Resolves #162.


flight-core/service.go
```diff
@@ -78,3 +106,15 @@ func (s *Service) mkPidfileDir() error {
...
+func processFromPidfile(pidfilePath string) (*os.Process, error) {
+	pid, err := pidfile.Read(pidfilePath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+	if pid == 0 {
+		// Process is no longer running.
+		return nil, nil
+	}
+	return os.FindProcess(pid)
```
benarmston: On Unix systems, FindProcess will always return a process (because Unix).  To test if there is actually a process we need to see if sending a signal to the process works or not. There is a special signal number that can used for that.  See https://github.com/alces-flight/flight-user-suite/blob/main/flight-core/pidfile/pidfile.go#L14 for details.

flight-core/service_start.go
```diff
@@ -17,17 +19,32 @@ func startCommand() *cli.Command {
...
+				pin.WithFailSymbol('\u274c'),
+				pin.WithFailColor(pin.ColorRed),
+			)
+			cancel := p.Start(ctx)
+			defer cancel()
+
+			// Pause for better spinner UX
+			<-time.After(1 * time.Second)
+
+			response, err := service.Start(ctx)
```
benarmston: A very minor and very pedantic improvement, so feel free to ignore. The code currently waits for 1 second plus the time taken for `service.Start` to return.  That could easily be changed to the greater of 1 second or the time taken for `service.Start` to return.

```suggestion
			timer := time.After(1 * time.Second)

			response, err := service.Start(ctx)
			// Pause for better spinner UX
			<- timer
```

flight-core/service_stop.go
```diff
@@ -15,12 +18,26 @@ func stopCommand() *cli.Command {
...
+				pin.WithFailSymbol('\u274c'),
+				pin.WithFailColor(pin.ColorRed),
+			)
+			cancel := p.Start(ctx)
+			defer cancel()
+
+			// Pause for better spinner UX
+			<-time.After(1 * time.Second)
+
 			err := service.Kill()
```
benarmston:
```suggestion
			timer := time.After(1 * time.Second)

			err := service.Kill()
			// Pause for better spinner UX
			<- timer
```

flight-web-suite/main.go
```diff
@@ -81,12 +85,36 @@ func main() {
...
+			successResponse := process.Response{
+				Success: true,
+				Message: fmt.Sprintf("Started Web Suite on %s", addr),
+			}
+			responseJson, _ := successResponse.ToJSON()
+
+			// The additional file handle passed from parent process will be ID 3
+			responseFile, err := os.OpenFile("/proc/self/fd/3", os.O_WRONLY, 0644)
+			if err != nil {
+				e.Logger.Error("failed to return response to calling process", "error", err)
```
benarmston: Is the file not existing an error worth logging? Or an expected use-case? If an expected use-case we could only log unexpected errors.

```suggestion
				if !strings.Contains(err.Error(), "no such device or address") {
				        e.Logger.Error("failed to return response to calling process", "error", err)
				}
```

flight-core/service_start.go
```diff
@@ -17,17 +19,32 @@ func startCommand() *cli.Command {
...
 			if err != nil {
-				return cli.Exit(fmt.Sprintf("Starting %s service failed: %s", service.Name, err), 1)
+				p.Fail(fmt.Sprintf("Starting %s service failed: %s", service.Name, err))
+				os.Exit(1)
+			}
+
+			if response.Success {
+				p.Stop(response.Message)
```
benarmston, line 45: We don't store the URL against the service, resulting in `flight web status` showing `-` for the URL. We ought to either store the URL in some way or remove the URL column from that table.

flight-web-suite/main.go
```diff
@@ -81,12 +85,36 @@ func main() {
...
+				fmt.Fprintln(responseFile, responseJson)
+				responseFile.Close()
+			}
+
+		},
+	}
+
+	if err := sc.Start(ctx, e); err != nil {
 		e.Logger.Error("failed to start server", "error", err)
+		os.Exit(1)
```
benarmston, line 118: We ought to be reporting errors here too.  If there is a process not started by `flight web start` (such as one started by `air`) listening on port the error shown to the user is

```
❌ Starting Web Suite service failed: Failed to read response from service process: EOF
```

We can improve that with something like. (I've not fully thought this through).

```suggestion
  		responseFile, openErr := os.OpenFile("/proc/self/fd/3", os.O_WRONLY, 0644)
		if openErr == nil {
			successResponse := process.Response{
				Success: false,
				Message: fmt.Sprintf("Failed to start Web Suite: %s", err),
			}
			responseJson, _ := successResponse.ToJSON()
			fmt.Printf("responseJson: %v\n", responseJson)
			_, _ = fmt.Fprintln(responseFile, responseJson)
			_ = responseFile.Close()
		}
		os.Exit(1)
```

Then we get

```
❌ Failed to start Web Suite: listen tcp 0.0.0.0:8080: bind: address already in use
```

benarmston (changes requested):
This is much better than it was before, but I think there is still a couple of issues that need to be looked at.  I've also added some super nitpicky suggestions.

jamesremuscat:
@benarmston I've addressed your comments. This is ready for another look (hopefully... :crossed_fingers: )

benarmston (approved):
Thanks for making those changes. LGTM.

PR #163, opened 2026-04-30
Improve sessions page layout

benarmston:
This PR adds a number of improvements to the session show page.

* Use more of the available width.
* Move some toolbar actions into a dropdown.
* Use heroicons.
* Better background.

image

PR #170, opened 2026-05-01
Support VNC fullscreen mode

### LINKED ISSUE #154: [Flight Web - Desktop] Support full-screen (and "zen mode"?) for viewing VNC sessions
As per title. We should at least support "full screen" mode. Optionally we could support a minimal-ui "Zen" mode.

jamesremuscat:
This PR adds the capability to set the VNC canvas to full-screen, via a new button on the toolbar. It is based on the branch under review in #163.

image

It behaves, hopefully, as you'd expect. The Stimulus controller also listens for the `fullscreenchange` event on `document` to detect when fullscreen mode has been exited outside of its own code.

Resolves #154 (at this point I think "zen mode" is something to implement iff we get asked for it specifically...)

### DIFF
--- a/flight-web-suite/assets/images/icons/arrows-pointing-out.svg
+++ b/flight-web-suite/assets/images/icons/arrows-pointing-out.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  fill="none"
+  viewBox="0 0 24 24"
+  stroke-width="1.5"
+  stroke="currentColor"
+  class="size-6"
+>
+  <path
+    stroke-linecap="round"
+    stroke-linejoin="round"
+    d="M3.75 3.75v4.5m0-4.5h4.5m-4.5 0L9 9M3.75 20.25v-4.5m0 4.5h4.5m-4.5 0L9 15M20.25 3.75h-4.5m4.5 0v4.5m0-4.5L15 9m5.25 11.25h-4.5m4.5 0v-4.5m0 4.5L15 15"
+  />
+</svg>
--- a/flight-web-suite/assets/javascript/application.js
+++ b/flight-web-suite/assets/javascript/application.js
@@ -5,12 +5,15 @@ import "./dropdown";
 // Import stimulus controllers. Each entry here needs manually registering
 // below.
 import DisableWithController from "./controllers/disable_with_controller.js"
-import NovncController from "./controllers/novnc_controller"
+import FullscreenController from "./controllers/fullscreen_controller.js"
 import MenuToggleController from "./controllers/menu_toggle_controller.js"
+import NovncController from "./controllers/novnc_controller"
 import RefreshController from "./controllers/refresh_controller.js"
 
 window.Stimulus = Application.start()
 Stimulus.register("disable-with", DisableWithController)
 Stimulus.register("novnc", NovncController)
+Stimulus.register("fullscreen", FullscreenController)
 Stimulus.register("menu-toggle", MenuToggleController)
 Stimulus.register("refresh", RefreshController)
+Stimulus.register("novnc", NovncController)
--- a/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js
+++ b/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js
@@ -1,0 +1,37 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["fullscreenable"]
+  static values = { isFullscreen: { type: Boolean, default: false } }
+
+  initialize() {
+    this.updateFromDocument = this.updateFromDocument.bind(this)
+  }
+
+  connect() {
+    document.addEventListener('fullscreenchange', this.updateFromDocument)
+  }
+
+  disconnect() {
+    document.removeEventListener('fullscreenchange', this.updateFromDocument)
+  }
+
+  isFullscreenValueChanged(isFullscreen) {
+    if (isFullscreen) {
+      this.fullscreenableTarget.requestFullscreen().catch(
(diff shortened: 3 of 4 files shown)


benarmston:
> Resolves #154 (at this point I think "zen mode" is something to implement iff we get asked for it specifically...)

Sounds sensible to me.

benarmston (commented):
There are a couple of issue's for me.

* When full screened, my cursor doesn't show inside the VNC session.
* When exiting full screen, the VNC session doesn't always return to the correct size.

The second of those is greatly improved, if I disable animations in my desktop, so it appears to be a timing issue.

jamesremuscat:
Do you have reproduction steps for either of those? I can't reproduce the issues in Chrome nor in Firefox.

benarmston:
> Do you have reproduction steps for either of those? I can't reproduce the issues in Chrome nor in Firefox.

Only the obvious. Make it full screen and observe the issue.  I've not done anything special to produce it.

I can produce both issues in Firefox, but only the cursor issue in Chromium.

benarmston:
The following diff fixes the cursor issue for me and greatly improves the resizing issue.  It changes the behaviour of the full screen button to be identical to pressing F11.

```diff
diff --git a/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js b/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js
index 7da4ce5..0c78e9d 100644
--- a/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js
+++ b/flight-web-suite/assets/javascript/controllers/fullscreen_controller.js
@@ -1,7 +1,6 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = ["fullscreenable"]
   static values = { isFullscreen: { type: Boolean, default: false } }
 
   initialize() {
@@ -18,16 +17,13 @@ export default class extends Controller {
 
   isFullscreenValueChanged(isFullscreen) {
     if (isFullscreen) {
-      this.fullscreenableTarget.requestFullscreen().catch(
-        () => {
+      document.documentElement.requestFullscreen().catch(
+        (err) => {
           // Fullscreening failed FSR
           this.isFullscreenValue = false
         }
       )
-    }
-    else if (this.fullscreenableTarget === document.fullscreenElement) {
-      // Not sure how this would be triggered given the fullscreen VNC canvas
-      // covers the button, but for completeness:
+    } else if (document.fullscreenElement != null) {
       document.exitFullscreen()
     }
   }
@@ -37,6 +33,6 @@ export default class extends Controller {
   }
 
   updateFromDocument() {
-    this.isFullscreenValue = this.fullscreenableTarget === document.fullscreenElement
+    this.isFullscreenValue = document.fullscreenElement != null
   }
 }
diff --git a/flight-web-suite/views/pages/desktop/show.gohtml b/flight-web-suite/views/pages/desktop/show.gohtml
index ee950a6..89c715f 100644
--- a/flight-web-suite/views/pages/desktop/show.gohtml
+++ b/flight-web-suite/views/pages/desktop/show.gohtml
@@ -65,7 +65,7 @@
     </div>
 
     <div class="h-[calc(100%_-_var(--spacing)_*_10)]">
-        <div class="novnc h-full" data-novnc-target="canvas" data-fullscreen-target="fullscreenable">
+        <div class="novnc h-full" data-novnc-target="canvas">
             <!-- This is where the desktop session will appear -->
         </div>
         <div style="display:none" data-novnc-target="copyFallback"></div>
```

jamesremuscat:
> It changes the behaviour of the full screen button to be identical to pressing F11

Except that pressing F11 helpfully _doesn't_ trigger `fullscreenchange` so it's really easy to get inconsistent state between the browser and `FullscreenController`.

I'm torn between wanting to investigate and understand what's going wrong in your browser, and scrapping this feature for time - I don't think the branch with the above patch adds enough functionality over the user just pressing F11 themselves to make it worth it.

benarmston:
> Except that pressing F11 helpfully _doesn't_ trigger `fullscreenchange` so it's really easy to get inconsistent state between the browser and `FullscreenController`.

I've not been able to do that.  Do you have reproduction steps?
 
> I'm torn between wanting to investigate and understand what's going wrong in your browser, and scrapping this feature for time - I don't think the branch with the above patch adds enough functionality over the user just pressing F11 themselves to make it worth it.

That's probably the most sensible thing to do if we're getting inconsistent behaviour.

jamesremuscat:
> I've not been able to do that. Do you have reproduction steps?

Load the show page and press F11 to go full-screen.

Now, instead of toggling full-screen, the full-screen button toggles between "To exit full screen press and hold Esc" and "To exit full screen press Esc" messages in Chrome but leaves the window full-screened!

benarmston:
> > I've not been able to do that. Do you have reproduction steps?
> 
> Load the show page and press F11 to go full-screen.
> 
> Now, instead of toggling full-screen, the full-screen button toggles between "To exit full screen press and hold Esc" and "To exit full screen press Esc" messages in Chrome but leaves the window full-screened!

I can now reproduce in both Firefox and Chromium.

jamesremuscat:
@benarmston I've rebased this branch on latest `main`; slightly clutching at straws but would you be able to see if the issues you were experiencing are still here?

Were you testing with a `gnome` session or an `xterm` one? If the former is it possible that something in that session was conflicting with your host laptop?

benarmston:
The mouse pointer issue still exists.  I could not reproduce the resize issue.  I've tested against running this locally and also against barely-a-cluster on garycloud.

jamesremuscat:
@megan-godfrey has tried this branch and hasn't experienced any issues, so I'm going to merge this through and include it in the release, and we can work out where your cursor has gone tomorrow @benarmston.

PR #171, opened 2026-05-05
[Flight Core] Capture service logs

benarmston:
Services, aka flight web suite, are expected to log to stdout. Flight core now redirects both stdout and stderr to
`/opt/flight/var/log/<service_id>.log`.

No log rotation has been implemented.


benarmston:
This is a small change.  Merging without review.

PR #174, opened 2026-05-05
Better VNC session launch feedback

### LINKED ISSUE #146: [Flight Web - Desktop] Better feedback on launching desktop session
Launching a desktop session can take several seconds (unusually, for a FWS HTTP request!).

We should improve the UX during this waiting time:
- Adding a spinner / disabling the "launch" button / changing the text to "launching" and prevent multiple clicks on the button
  - Though I do note that in-flight requests are cancelled if you spam-click the button; I'm not sure how thoroughly we clean up those cancelled requests though?
- Maybe highlighting the new session in some way when we go back to the index page? (Or maybe, when we have session viewing, we won't need to?)

jamesremuscat:
This PR makes a couple of improvements to VNC session launching via Flight Web Suite.

- When the launch form is submitted, change the text of the submit button and disable it to avoid multiple submissions and to provide visual feedback that something has happened.
  image

- Redirect to the show page for the new session, not the session index page.
 
I've deliberately followed Rails-y conventions for the button disabling, even though that makes the data attributes inconsistent with Stimulus' standard practice. Having a `data-disable-with` attribute just feels more familiar than `data-disable-with-caption-value` or similar!

Resolves #146.

### DIFF
--- a/flight-web-suite/assets/javascript/application.js
+++ b/flight-web-suite/assets/javascript/application.js
@@ -4,11 +4,13 @@ import "./dropdown";
 
 // Import stimulus controllers. Each entry here needs manually registering
 // below.
+import DisableWithController from "./controllers/disable_with_controller.js"
 import NovncController from "./controllers/novnc_controller"
 import MenuToggleController from "./controllers/menu_toggle_controller.js"
 import RefreshController from "./controllers/refresh_controller.js"
 
 window.Stimulus = Application.start()
+Stimulus.register("disable-with", DisableWithController)
 Stimulus.register("novnc", NovncController)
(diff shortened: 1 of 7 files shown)


benarmston (approved):
This is much nicer.

I did run into an issue where the show page was sometimes rendered before the websocket server was ready. This resulted in the VNC being stuck in the disconnected state until the user clicked on the Connect button/placeholder image.

I've pushed a commit to address that.  If the VNC session is disconnected uncleanly, a single attempt is made to reconnect after 500ms.  This addresses the immediate issue without resulting in continuous attempts to reconnect.

The manual disconnect button still works as it should.

With that commit in place this LGTM.

PR #175, opened 2026-05-06
Add screenshots to desktop index page

### LINKED ISSUE #158: [Flight Web - Desktop] Support grabbing and storing screenshots from sessions
The design reference included screenshots of running sessions:

Image

We should create a mechanism to capture a screenshot and store it within the session directory for use in this way. (How this could work is open to interpretation - happy for this to only work with sessions that have been viewed within FWS if that makes it easier)

Bonus points to also create a mechanism for the end user to capture screenshots and save them locally to their device.

benarmston:
The images are shown at a fixed height and max width. If the image is wider than allowed at the fixed height, it is centered and clipped.

image

I've also double checked that images are correctly captured on a "barley-a-cluster" cluster running on garycloud.  I think the setup their is closer to the setup we need to support than our laptops.

Closes #158

### DIFF
--- a/flight-web-suite/views/pages/desktop/index.gohtml
+++ b/flight-web-suite/views/pages/desktop/index.gohtml
@@ -57,13 +57,29 @@
                     </div>
                 </div>
             </div>
-            <dl class="details-grid grid grid-cols-[auto_1fr] gap-x-6 gap-y-2">
+            <a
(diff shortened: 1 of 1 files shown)


megan-godfrey (approved):
LGTM!

PR #176, opened 2026-05-06
Add `webify` command

### LINKED ISSUE #159: [Flight Web - Desktop] Support starting `websockify` for existing running sessions
Add support for starting `websockify` after the fact. If a session has been started before `websockify` has been installed, the VNC session will be running without `websockify`. We'd like to be able to start `websockify` by running, say, `flight desktop webify <sessionName>`.

(In practice not sure how useful this will be - so lower priority than other enhancements.)

benarmston:
This PR adds a `webify` command to `flight-desktop`.

This allows the web support services for older sessions to be manually started. No attempt has been made to have web suite automate this process.  Whilst that would be really useful, it is not as straight forward as implementing the `webify` command.

The `webify` command starts the websockify process, the screenshot grabber script and the cleaner script.  A check is made around websockify and screenshot grabber to ensure that they are not started if already running. If
either websockify or grabber are started a new cleaner script is started.

If a new cleaner script is started, we probably ought to kill the old cleaner script. However, there is a maximum of three cleaner scripts running per session, It spends most of its time sleeping and having multiple run at once cannot interfere with each other.

### Future enhancements

* Web suite should automatically run the `webify` command when needed.  Perhaps, the desktop index page, could check a new attribute `IsWebified` on each session and run `webify` for each session that has not been webified.  `IsWebified` would be true if the session has both a running websockify and grabber process and false otherwise.

Closes #159

### DIFF
--- a/flight-desktop/session_webify.go
+++ b/flight-desktop/session_webify.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"charm.land/log/v2"
+	"github.com/muesli/reflow/wordwrap"
+	"github.com/urfave/cli/v3"
+)
+
+func webifySessionCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "webify",
+		Usage:       "Start web access support for an active desktop session",
+		Description: wordwrap.String("Start required and optional web support programs for an active interactive desktop session.", maxTextWidth),
+		Category:    "Sessions",
+		// Flags:       []cli.Flag{formatFlag},
+		Arguments: []cli.Argument{
+			&cli.StringArg{Name: "name", UsageText: "<name>"},
+		},
+		Before:        assertArgPresent("name"),
+		ShellComplete: completeActiveSessionNames,
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			name := cmd.StringArg("name")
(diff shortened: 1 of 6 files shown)


flight-desktop/session_webify.go
```diff
@@ -0,0 +1,59 @@
...
+	"charm.land/log/v2"
+	"github.com/muesli/reflow/wordwrap"
+	"github.com/urfave/cli/v3"
+)
+
+func webifySessionCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "webify",
+		Usage:       "Start web access support for an active desktop session",
+		Description: wordwrap.String("Start required and optional web support programs for an active interactive desktop session.", maxTextWidth),
```
jamesremuscat: Not sure if here is the best place or not, but it might be useful to explain _why_ this command might be useful (and that it's not necessary to manually run it for every session).

Something like (feel free to tweak):

```suggestion
		Description: wordwrap.String("Start required and optional web support programs for an active interactive desktop session. Can be used to add web support to sessions that were started before web dependencies were installed.", maxTextWidth),
```

jamesremuscat (changes requested):
LGTM, just one suggestion to consider.

benarmston:
Suggestion committed.  Merging this without further review.

PR #178, opened 2026-05-06
Add cache buster to screenshots on page load

### LINKED ISSUE #177: [Flight Web - Desktop] Fix caching issues with desktop screenshots
There is a caching issue with desktop screenshots.

I suspect that, on page load the URL does not contain a cache busting query parameter.  When the screenshot refreshes the cache busting param is added, but prior to that the screenshot might be cached as the placeholder or the initial screenshot from before any applications were opened.

benarmston:
Previously, the cache buster was only added once the screenshot was refreshed after 5 minutes.  Until then it would potentially use a cached value which may well have been the placeholder image.

Fixes #177

### DIFF
--- a/flight-web-suite/assets/javascript/controllers/refresh_controller.js
+++ b/flight-web-suite/assets/javascript/controllers/refresh_controller.js
@@ -4,19 +4,33 @@ import { Controller } from "@hotwired/stimulus"
 //
 // Connects to data-controller="refresh"
 export default class extends Controller {
-  static values = { interval: { type: Number, default: 60 * 1000 } }
+  static values = {
+    interval: { type: Number, default: 60 * 1000 },
+    src: String,
+  }
+
+  initialize() {
+    this.refresh = this.refresh.bind(this)
+  }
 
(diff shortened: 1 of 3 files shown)


jamesremuscat (approved):
This works - is there a need to special-case the initial image load, or could we simplify by just calling `refresh()`  in `srcValueChanged`?

Happy for this to be merged either way.

benarmston:
> This works - is there a need to special-case the initial image load, or could we simplify by just calling `refresh()` in `srcValueChanged`?

The special casing adds the non-cache busted src ASAP.  There's a good chance that this has been cached, so it appears on the page without "flickering" from no-image to an image.  It is then swapped for the correct image, but as the images are displayed at the same height and width, the flickering is minimal.

Simply calling `refresh()` in `srcValueChanged` would skip that "flicker" minimising routine.

An alternative would be to encode the placeholder image as a data url. Display that ASAP and then fetch and replace it with the cache busted image.  But I thought of that after having implemented the above and decided what's there is good enough.

PR #179, opened 2026-05-06
Change definition of admin user

benarmston:
Previously, we defined the admin user as `root`.  Now, we define the admin user as either `root` or user with password-less `sudo` access.

This makes using FUS for admin users a little nicer as they don't have to prefix each command with `sudo`. It also, results in the admin howto guides being displayed in the FWS, which is the biggest win.

The implementation consists of wrapping each admin command with `userroles.AsRoot`. `AsRoot` checks if the program is running as the root user and if so directly invokes the wrapped function.  Otherwise, `AsRoot` re-executes the command prefixed with `sudo --preserve-env=FLIGHT_ROOT,FLIGHT_STATE_ROOT`.

For listing the howto guides, checking if the current user is `root` to filter the admin guides, has been replaced with checking if the user `IsAdmin`.

image

### Testing

To test this in the terminal run `sudo /bin/true` and enter your password.  You will then have password-less `sudo` access until it expires.  You can force its expiration with `sudo -k`.  To test this for the web suite, edit `/etc/sudoders` and adjust to make your account have password-less sudo access.  You'll want to remember to undo that change.

### Future enhancements

* Improve the styling for the list of howto guides. Each "section", e.g., "Admin Guides" should become a header and the guides under that header have the prefixed removed.
* Consider how we want to handle admins with password guarded `sudo` access.  We could _potentially_ detect this and consider them an admin when running in an interactive terminal.
* Consider how we want to handle users with password-less `sudo` access, but not password-less access to run `flight`.
 
The last two would be nice to deal with, but the original FUS didn't, likely because that's not a common setup on our clusters.  They are both "safe" as they currently exist: admins can run `sudo flight ...` and enter their password; re-execing for not-actually-admins will fail.

### DIFF
--- a/flight-core/config_command.go
+++ b/flight-core/config_command.go
@@ -15,6 +15,7 @@ import (
 	"charm.land/log/v2"
 	"github.com/adrg/xdg"
 	"github.com/concertim/flight-user-suite/flight/cliui"
+	"github.com/concertim/flight-user-suite/flight/userroles"
 	"github.com/hashicorp/go-envparse"
 	"github.com/muesli/reflow/wordwrap"
 	"github.com/urfave/cli/v3"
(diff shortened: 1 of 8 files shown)


jamesremuscat (approved):
This LGTM. It kinda-sorta feels like magic; I run `sudo` for some other command and then I magically get all the admin options!

PR #180, opened 2026-05-06
Support custom screen geometry

### LINKED ISSUE #148: [Flight Web - Desktop] Support user-specified desktop geometries
At the moment (#138) we hard-code a couple of desktop geometries.

Design reference included the ability for users to specify their own geometries; we should allow them to do so, subject to any limits imposed by the host system (if any - is there a maximum size specified somewhere?)

We'll also need to consider scaling or scrolling the NoVNC canvas if the desktop resolution is larger than can fit in the viewport.

jamesremuscat:
This PR adds the ability to specify custom screen geometry to the desktop launch form. It also adds several more hard-coded geometry options to the dropdown menu.

image

The custom resolution fields are only shown when "custom" is selected from the dropdown; they are also marked as `required` in that scenario.

`disable_with_controller` has become smarter about disabling a button within a form; it waits for the form's (post-validation) `submit` event before disabling the button. Otherwise, trying to submit an invalid form would disable the button with no way of un-disabling it once the validation errors had been corrected.

Resolves #148.

### DIFF
--- a/flight-web-suite/assets/javascript/application.js
+++ b/flight-web-suite/assets/javascript/application.js
@@ -6,12 +6,14 @@ import "./dropdown";
 // below.
 import DisableWithController from "./controllers/disable_with_controller.js"
 import FullscreenController from "./controllers/fullscreen_controller.js"
+import GeometryInputsController from "./controllers/geometry_inputs_controller.js"
 import MenuToggleController from "./controllers/menu_toggle_controller.js"
 import NovncController from "./controllers/novnc_controller"
 import RefreshController from "./controllers/refresh_controller.js"
 
 window.Stimulus = Application.start()
 Stimulus.register("disable-with", DisableWithController)
+Stimulus.register("geometry-inputs", GeometryInputsController)
 Stimulus.register("novnc", NovncController)
 Stimulus.register("fullscreen", FullscreenController)
 Stimulus.register("menu-toggle", MenuToggleController)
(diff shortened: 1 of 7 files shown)


benarmston (commented):
The functionality works and the UI/UX is nice in the good case. However, there are some error handling issues to look at.

(Github isn't accepting inline comments at the moment.  So everything is being added here (hopefully).)

* The event listeners added in `disable_with_controller` ought to be removed when the controller disconnects.
* The handling around re-rendering the form after an invalid submission isn't quite right.

To reproduce:

* Enter bad name and good custom resolution:
  Image
* Submit the form, notice that "custom" is selected, but the inputs are not shown.
  Image
* Change away from custom and back to custom to show the custom inputs. Notice that they are not pre-populated with the valid values entered previously:
  Image

benarmston (changes requested):

jamesremuscat:
> The event listeners added in disable_with_controller ought to be removed when the controller disconnects.

Strictly speaking, that's true, though it's less of a concern here outside Turbo's document morphing - if the controller is disconnecting then it's because the page is reloading or being replaced with a new page.

It's awkward to remove the handler once the button has been removed from the DOM as I don't think we can rely on `button.form` pointing to anything at that point...

jamesremuscat:
@benarmston I've addressed your other comments; good catch.

benarmston (approved):
I've pushed a commit to address an issue with re-rendering the form after a submission failure.  With that included, this LGTM.

PR #184, opened 2026-05-07
[Web Suite - Desktop] Make multiple reconnection attempts

benarmston:
If a connection to a websockify process fails, make a number of attempts to reconnect with increasing backoff.

Twelve attempts will be made over some thirty-odd seconds. If the session is still not connected after that, no more attempts will be made until the user clicks to connect.


benarmston:
@jamesremuscat do you want to take a look at this before I merge?

jamesremuscat (approved):
LGTM.

PR #185, opened 2026-05-07
Remove false positive admin checks

benarmston:
Admin check gives almost(?) zero false positives

When checking if a user should be considered an admin, we now consider if they have password-less `sudo` access to run `flight`, instead of any kind of password-less sudo.

A user could be setup with password-less `sudo` access to run a single specific command, but not general `sudo` access.  Checking against `flight` prevents that false positive.

We have to determine the absolute path to `flight` when actually running `sudo`, so we can use the same mechanism to determine the path to make the sudo check against.

This will work for production setups that grant specific sudo access, e.g., just to run `/opt/flight/bin/flight`; or for production setups that provide general sudo access to everything.

In development, it is a little more awkward, but still manageable. Password-less sudo access could be setup to run the binary in `dist`, e.g., `~/projects/flight-user-suite/dist/opt/flight/bin/flight`, which will then work nicely with web suite. Alternatively, if general sudo access is available, `go run` will detect the user as an admin.

### DIFF
--- a/flight-core/opt/flight/usr/share/doc/flight-user-suite/01-about-flight-user-suite.md
+++ b/flight-core/opt/flight/usr/share/doc/flight-user-suite/01-about-flight-user-suite.md
@@ -60,5 +60,5 @@ running `flight howto list` in an active flight environment.
 
 Flight User Suite considers a user an admin if they:
 - Are `root`
-- Have passwordless sudo (which includes if sudo is currently passwordless due
-  to having a cached password)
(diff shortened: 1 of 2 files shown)


jamesremuscat (commented):
LGTM - but `main` has a documentation change that's invalidated by these changes and will need to be updated (99998c79d3567228264c63f026e4f448ee533dfc is the original commit).

benarmston:
Rebased on `main` and documentation addressed.

PR #188, opened 2026-05-07
Add `web doctor` command to check web suite dependencies

### LINKED ISSUE #164: [Flight Web] Add `web doctor` command to check dependencies
FWS has a dependency on both Python and (the correct) `python-pam`.

We should introduce a `web doctor` command (analog of `desktop doctor`) to check these dependencies are in place.

benarmston:
This PR adds support for checking web suite dependencies via a `flight web doctor` command.

The checking of desktop web access dependencies has remained in `desktop doctor` so that it can be ran on a variety of machines.  Running `desktop doctor` on a visualisation node where desktop sessions might be started, seems to make more sense to me than running `web doctor` on a node where FWS will never run.

Checking desktop web access dependencies could have been duplicated in `web doctor`.  However, the machine that FWS is to run on might not be intended as a desktop target and might not have desktop enabled on that machine or its dependencies installed.

On balance, the better solution is to have `web doctor` only check for general (aka core) dependencies.

```bash
$ flight web doctor

✅ Required general Flight Web Suite dependencies

 > ✅ Python 3   : /usr/bin/python3            
 > ✅ Python PAM : /usr/bin/python3 imports pam

✅ All required dependencies satisfied
```

If the checks fail, it is expected that the user will read the find documentation to understand what needs to be installed.

### Implementation

* The doctor related routes and types have been extracted to a new `doctor` package and `flight-desktop` updated.
* A new dependency type `python_module` has been added.  It checks that `python3 -c 'import <module>'` exits successfully.
* A new `web doctor` command has been added.  The dependencies it checks are hardcoded.  I see no point in allowing them to be configurable, when the `authenticate.py` script effectively hardcodes them too.

### DIFF
--- a/flight-core/opt/flight/usr/share/doc/flight-web-suite/99-admin-guides/30-flight-web.md
+++ b/flight-core/opt/flight/usr/share/doc/flight-web-suite/99-admin-guides/30-flight-web.md
@@ -15,6 +15,15 @@ Flight Web Suite requires:
 * The [python-pam](https://pypi.org/project/python-pam/) library. Known as
   `python3-pampy` on Ubuntu. Known as `python3-pam` on Rocky 9 available from
   EPEL.
+
+You can verify these prerequisites with:
+
+```bash
+flight web doctor
+```
+
+If the Flight desktop tool is enabled:
+
 * `/usr/bin/websockify` to provide browser access to Flight Desktop sessions
 * Optionally, `/usr/bin/import` from ImageMagick to enable screenshot capture
   for Flight Desktop sessions
@@ -22,7 +31,7 @@ Flight Web Suite requires:
 You can verify these prerequisites with:
 
 ```bash
-flight web doctor
+flight desktop doctor
 ```
 
 ## Configuration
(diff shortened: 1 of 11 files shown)


flight-core/opt/flight/usr/share/doc/flight-web-suite/99-admin-guides/30-flight-web.md
```diff
@@ -49,6 +67,12 @@ deployment configuration.
   flight web status
   ```
 
+* **Check dependencies:** as `root` run
```
jamesremuscat: IIUC this doesn't need to be run as `root` but rather as an admin (according to FWS's definition)?

It certainly _works_ for me as non-root with passwordless sudo.

jamesremuscat (commented):
This LGTM, just one query on some of the documentation.

flight-core/opt/flight/usr/share/doc/flight-web-suite/99-admin-guides/30-flight-web.md
benarmston: That sounds correct. Would changing this to



```suggestion
* **Check dependencies:** run the following as a cluster admin user
```

jamesremuscat: That wording sounds fine to me.

PR #193, opened 2026-05-11
Improved remote session support

### LINKED ISSUE #190: [Flight Web - Desktop] Improve support for remote sessions
* Detect if a remote session is correctly webified.
* Terminate remote sessions.
* Clean up remote sessions.
* Webify both local and remote sessions as needed.

benarmston:
Add support for

* [x] Terminating and cleaning up remote sessions
* [x] Webifying both local and remote sessions
* [ ] Write PR description


flight-web-suite/cli/run.go
```diff
@@ -0,0 +1,143 @@
...
+		Auth: []ssh.AuthMethod{publicKey},
+
+		// TODO: Do we need to change this? We create an SSH keypair via the
+		// ssh-keypair-generation hook and configure the ssh config with
+		// `StrictHostKeyChecking no`. Would changing this just be a pretence?
+		//
+		// hostKeyCallback, err := knownhosts.New(filepath.Join(homedir, ".ssh", "known_hosts"))
+		// if err != nil {
+		// 	return nil, cleanup, fmt.Errorf("creating hostKeyCallback: %w", err)
+		// }
```
jamesremuscat, line 116: In an ideal world, we would honour whatever is set in the user's `.ssh/config` (it would be counterintuitive, for me at least, to have something operating `ssh` on my behalf and to ignore that config file).

In a more pragmatic world, if FUS is setting an SSH config, and the config here is consistent with that config, I don't see any issues.

jamesremuscat (approved):
I've not fully tested these changes, but the code changes make sense to me. I've given an answer to your inline question.

